PR #3, opened 2018-01-27
CLI client!

tgoodlet:
Full api via cli support!

### DIFF
--- a/piker/brokers/cli.py
+++ b/piker/brokers/cli.py
@@ -1,0 +1,29 @@
+"""
+Console interface to broker client/daemons.
+"""
+from pprint import pformat
+import click
+import trio
+from ..log import get_console_log
+
+
+def run(loglevel, main):
+    log = get_console_log(loglevel)
+
+    # main loop
+    try:
+        client = trio.run(main)
+    except Exception as err:
+        log.exception(err)
+    else:
+        log.debug(
+            f"Exiting with last access info:\n{pformat(client.access_data)}\n")
+
+
+@click.command()
+@click.option('--broker', default='questrade', help='Broker backend to use')
+@click.option('--loglevel', '-l', default='warning', help='Logging level')
+def pikerd(broker, loglevel):
+    # import broker module daemon entry point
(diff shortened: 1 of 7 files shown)


piker/brokers/cli.py
```diff
@@ -0,0 +1,61 @@
...
+    finally:
+        log.debug("Exiting piker")
+
+
+@click.command()
+@click.option('--broker', default='questrade', help='Broker backend to use')
+@click.option('--loglevel', '-l', default='info', help='Logging level')
+def pikerd(broker, loglevel):
+    # import broker module daemon entry point
+    brokermod = import_module('.' + broker, 'piker.brokers')
```
vodik: Have you considered using entry points for this? http://amir.rachum.com/blog/2017/07/28/python-entry-points/

Has some good benifits if you want to make either pieces optional (entry points can be bound with optional dependencies, handling that for you) or if you want to allow other people to add their own support. I've used this for some services and it makes developing stuff out of tree easy.

Granted if its all going to be in tree, its a different story and maybe a bit more complicated than its worth.

piker/brokers/cli.py
```diff
@@ -0,0 +1,61 @@
...
+def run(main, loglevel='info'):
+    log = get_console_log(loglevel)
+
+    # main sandwich
+    try:
+        return trio.run(main)
+    except Exception as err:
+        log.exception(err)
+    finally:
+        log.debug("Exiting piker")
```
vodik: I'm not crazy in that this code isn't being used yet?

piker/brokers/cli.py
tgoodlet: I'll look but for now I think since there's only one backend I'll keep everything together for the moment.

piker/brokers/cli.py
vodik: This one, the main function here.

tgoodlet: It's used on lines `32` and `61`?

vodik: Derp

PR #31, opened 2018-03-31
Tolerate the network

tgoodlet:
Adds support for roaming on wifi such that losing the network and later getting it back resumes all real-time data feeds.

Also moved some qt specific stuff into that broker mod.

@Konstantine00 @vodik @dkeitel


Konstantine00:
Looks good to me, wish I could provide useful feedback. >_<

PR #33, opened 2018-04-10
Watchlist builtins

tgoodlet:
Move the built-ins watchlists to a module variable. I also adjusted some things in `piker.watchlists` mostly to raise errors when removing non-existing entries.  Adjusted the tests to match.

Oh and I added a little extra logic to only report when the network *first* goes down and then comes back up instead of repeatedly spamming the same message to the console.

@Konstantine00 please give this a quick review if you can as it touches much of what you worked on.

### DIFF
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -112,7 +112,7 @@ def piker_dir(temp_dir):
         'pharma': ['ATE.VN'],
         'indexes': ['SPY', 'DAX', 'QQQ', 'DIA'],
     }
-    wl.write_sorted_json(watchlist, json_file_path)
+    wl.write_to_file(watchlist, json_file_path)
     yield json_file_path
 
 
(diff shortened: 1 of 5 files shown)


tests/test_cli.py
```diff
@@ -104,71 +101,77 @@ def temp_dir():
...
 
 @pytest.fixture
-def piker_dir(temp_dir):
-    wl.make_config_dir(temp_dir)
-    json_file_path = os.path.join(temp_dir, 'watchlists.json')
-    watchlist = {
-        'dad': ['GM', 'TSLA', 'DOL.TO', 'CIM', 'SPY', 'SHOP.TO'],
+def ex_watchlists():
+    """Made up watchliststo use for expected outputs.
+    """
```
Konstantine00: typo in the comments.

tgoodlet:
Bahh why not, I'm gonna add `TravisCI` before this gets in.

PR #36, opened 2018-04-16
Daemonize broker quote engine

### LINKED ISSUE #35: Daemonize quote streamers
As part of moving working towards #27 I'd like to start by initially running broker backend(s) in a separate process where client apps subscribe for and receive data feeds over IPC. For now I want to get this working with TCP sockets but eventually we'll likely move to ØMQ once we build out `trio` support for it.

This will also make long running backend code that much easier to [system test](https://en.wikipedia.org/wiki/System_testing) particularly when it comes to handling subscription patterns and reliability under different faults (network outages for ex).

tgoodlet:
Enables running the broker backends quote engine in a separate process and having clients subscribe for price streams wrapped in a queue api.

Still a WIP and I'll likely iterate on this a bit more but I wanted to get some eyes on it including:
- [x] making subscriptions per-broker (i.e. ticker lists should be broker specific for now)
- [x] using [`msgpack`](https://github.com/msgpack/msgpack-python) instead of `json`
- [x] spin up the daemon in a `multiprocessing.Process` if one is not already running and a client is launched? Maybe the same if the daemon crashes?
- [x] update docs showing how to launch the daemon and possibly print quotes to the console
- ~[ ] add a IPC way to tear down `trio.run()` launched in a subprocess~

Leave the last one. I think it's sensible to leave the child process daemon up when an app is torn down in case other apps still have live connections. Still need to figure out why the `SIGINT` isn't graceful in the subproc though...

Resolves #35

@Konstantine00 hint hint

### DIFF
--- a/piker/brokers/core.py
+++ b/piker/brokers/core.py
@@ -12,6 +12,7 @@ from typing import AsyncContextManager
 import trio
 
 from ..log import get_logger
+from . import get_brokermod
 log = get_logger('broker.core')
 
 
@@ -126,10 +127,8 @@ class StreamQueue:
 
 
 async def poll_tickers(
-    client: 'Client',
-    quoter: AsyncContextManager,
-    tickers: [str],
-    queue: StreamQueue,
+    brokermod: ModuleType,
+    tickers2qs: {str: StreamQueue},
     rate: int = 5,  # delay between quote requests
     diff_cached: bool = True,  # only deliver "new" quotes to the queue
 ) -> None:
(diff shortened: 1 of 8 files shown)


tgoodlet:
gd, typo in be7defde4646edb44dba97f6c8a5093099eacca5 message...

tgoodlet:
Needs one more commit and a rebase then I'll likely land it.
If you want to learn a load of stuff by reading code @Konstantine00 this is your chance.

piker/brokers/core.py
```diff
@@ -67,11 +69,132 @@
...
+            return await self.squeue.get()
+        except trio.BrokenStreamError as err:
+            if self._autorecon:
+                await self._reconnect()
+                return await self.recv()
+
+    async def aclose(self, *args):
+        await self.squeue.stream.aclose()
+
+    async def __aenter__(self):
```
tgoodlet: Hmm I think this should take in an optional named argument `sockaddr=None` which is passed through to `connect()`.

piker/brokers/core.py
```diff
@@ -67,11 +69,132 @@
...
+        while True:
+            try:
+                with trio.move_on_after(3) as cancel_scope:
+                    await self.connect()
+                cancelled = cancel_scope.cancelled_caught
+                if cancelled:
+                    log.warn("Reconnect timed out after 3 seconds, retrying...")
+                    continue
+                else:
+                    log.warn("Stream connection re-established!")
```
tgoodlet: Should this be `warning` level?

piker/brokers/core.py
```diff
@@ -67,11 +69,132 @@
...
+                    break
+            except (OSError, ConnectionRefusedError):
+                if not down:
+                    down = True
+                    log.warn(
+                        f"Connection to {self._sockaddr} went down, waiting"
+                        " for re-establishment")
+                await trio.sleep(1)
```
tgoodlet: Should the sleep time be configurable?

piker/brokers/core.py
```diff
@@ -67,11 +69,132 @@
...
+        """
+        try:
+            async for item in self.squeue:
+                yield item
+        except trio.BrokenStreamError as err:
+            if not self._autorecon:
+                raise
+        if self._autorecon:  # attempt reconnect
+            await self._reconnect()
+            async for item in self.aiter_recv():
```
tgoodlet: Hmm every time there's a broken stream we may be adding one layer to the call/iteration stack.
I wonder is there a more cleanly way to do the same?

piker/brokers/core.py
```diff
@@ -81,54 +204,198 @@
...
-            postquote_start = time.time()
+        cancelled = cancel_scope.cancelled_caught
+        if cancelled:
+            log.warn("Quote query timed out after 3 seconds, retrying...")
+            # handle network outages by idling until response is received
+            quotes = await wait_for_network(partial(get_quotes, tickers))
+
+        postquote_start = time.time()
+        q_payloads = {}
+        for symbol, quote in quotes.items():
```
tgoodlet: I think maybe this can be removed now? Might want to wait until we have tests.

piker/brokers/core.py
```diff
@@ -81,54 +204,198 @@
...
+) -> None:
+    """Handle per-broker quote stream subscriptions.
+
+    Spawns new quoter tasks for each broker backend on-demand.
+    Since most brokers seems to support batch quote requests we
+    limit to one task per process for now.
+    """
+    queue = StreamQueue(stream)  # wrap in a shabby queue-like api
+    log.info(f"Accepted new connection from {queue.peer}")
+    async with queue.stream:
```
tgoodlet: Hmm does this drop the connection with the client if all subscriptions are removed?
Is that desired?

piker/brokers/core.py
```diff
@@ -81,54 +204,198 @@
...
+        async for broker, tickers in queue:
+            log.info(
+                f"{queue.peer} subscribed to {broker} for tickers {tickers}")
+
+            if broker not in broker2tickersubs:
+                brokermod = get_brokermod(broker)
+                log.info(f"Spawning quote streamer for broker {broker}")
```
tgoodlet: drop this..

piker/brokers/core.py
```diff
@@ -81,54 +204,198 @@
...
+    limit to one task per process for now.
+    """
+    queue = StreamQueue(stream)  # wrap in a shabby queue-like api
+    log.info(f"Accepted new connection from {queue.peer}")
+    async with queue.stream:
+        async for broker, tickers in queue:
+            log.info(
+                f"{queue.peer} subscribed to {broker} for tickers {tickers}")
+
+            if broker not in broker2tickersubs:
```
tgoodlet: Hmm can the `clients` `dict` be used instead of this?

piker/brokers/core.py
```diff
@@ -81,54 +204,198 @@
...
+                get_quotes = await brokermod.quoter(client, tickers)
+                clients[broker] = (
+                    brokermod, client, client_cntxmng, get_quotes)
+                tickers2qs = broker2tickersubs.setdefault(
+                    broker, {}.fromkeys(tickers, {queue,}))
+            else:
+                log.info(f"Subscribing with existing `{broker}` daemon")
+                brokermod, client, _, get_quotes = clients[broker]
+                tickers2qs = broker2tickersubs[broker]
+                # update map from each symbol to requesting new client's queue
```
tgoodlet: *requesting* or *new*, which is it..

piker/cli.py
```diff
@@ -115,20 +126,61 @@ def quote(loglevel, broker, tickers, df_output):
...
     watchlist_from_file = wl.ensure_watchlists(_watchlists_data_path)
     watchlists = wl.merge_watchlist(watchlist_from_file, wl._builtins)
-    broker_limit = getattr(brokermod, '_rate_limit', float('inf'))
-
-    if broker_limit < rate:
-        rate = broker_limit
-        log.warn(f"Limiting {brokermod.__name__} query rate to {rate}/sec")
+    tickers = watchlists[name]
+
+    # setup ticker stream
```
tgoodlet: This comment isn't really necessary and the import can probably go top-of-file.

piker/cli.py
```diff
@@ -115,20 +126,61 @@ def quote(loglevel, broker, tickers, df_output):
...
+
+    async def main(timeout=1):
+        async def subscribe(client):
+            # initial request for symbols price streams
+            await client.send((brokermod.name, tickers))
+
+        client = Client(('127.0.0.1', 1616), subscribe)
+        start = time.time()
+        down = False
+        while True:
```
tgoodlet: Hmm I wonder if this wait loop is even necessary? The race for the `piker` port should be handled by the OS and spinning here I think does very little.

piker/ui/watchlist.py
```diff
@@ -372,6 +369,10 @@ def color_row(row, data):
             color_row(row, record)
 
         grid.render_rows(cache)
+        log.debug("Waiting on quotes")
+
+    log.warn("Server connection dropped")
+    nursery.cancel_scope.cancel()
```
tgoodlet: Hmm with the reconnect handling does this even get hit any more?

piker/ui/watchlist.py
```diff
@@ -381,82 +382,78 @@ def color_row(row, data):
...
+
+    async with trio.open_nursery() as nursery:
+        # get first quotes response
+        log.debug("Waiting on first quote...")
+        quotes = await client.recv()
+        first_quotes = [
+            brokermod.format_quote(quote, symbol_data=sd)[0]
+            for quote in quotes.values()]
+
+        if first_quotes[0].get('last') is None:
```
tgoodlet: Should go in the backend...

piker/brokers/core.py
tgoodlet: scratch that, [`__enter__` doesn't allow that](https://docs.python.org/3.6/reference/datamodel.html#object.__enter__)...

piker/cli.py
tgoodlet: switched.

piker/ui/watchlist.py
tgoodlet: Actually no this is fine for now.

piker/brokers/core.py
tgoodlet: nahh it's messy and aggregating to many refs together.

tgoodlet:
@Konstantine00 one last call for looking at this. Merging it this aft.

Issue #27, opened 2018-03-28
Distributed architecture model


tgoodlet:
I'd like to move to a distributed architecture where broker-daemons, clients / UI components and algorithmic and analytic frameworks run in separate processes with zero state sharing.

Some models I've been investigating include:
- the [actor model](https://en.wikipedia.org/wiki/Actor_model) for which [`pulsar`](http://quantmind.github.io/pulsar/design.html) serves as a good `asyncio` based reference solution
- a [multi-agent system](https://en.wikipedia.org/wiki/Multi-agent_system) for which [`aiomas`](https://aiomas.readthedocs.io/en/latest/guides/agent.html) appears to be a good `asyncio` based reference solution

Since I'd like to stick with the up and coming `trio` ecosystem, whatever solution is chosen will likely have to be (re-)written with that support.

tgoodlet:
Further discussion and brainstorming can be moved to my new project (which solves this issue): [tractor](https://github.com/tgoodlet/tractor)!

PR #50, opened 2018-06-07
Introducing "tractor"

tgoodlet:
Introduces a micro-framework for `multiprocessing` + `trio` based concurrency using an [actor model](https://en.wikipedia.org/wiki/Actor_model). 

This is super young and thus very rough and there's many ideas I need to work out including but not limited to:

- [x] actor nurseries (based on `trio`'s own `nurseries`)
- [x] actor cancellation via `Channel` (should work across OSes)
- [x] arbiter registry management, local execution and and per host scope rules
- [x] a `Channel` transparent async generator api for remote task synchronization
 -- this is mostly there except for support for `asend()` across channels; need some discussion with @vodik likely to think that out)

I'll likely make more issues to address these topic moving forward though not sure If it'll be in a new repo yet.

@vodik @Konstantine00 if you two don't mind giving this a peak I'd be super appreciative!

Relates to #27

### DIFF
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -1,7 +1,7 @@
 {
     "_meta": {
         "hash": {
-            "sha256": "e3eb385c3342f53cdb1793dc2055ec8f42669a15854b358632a5ceaa9dc1135b"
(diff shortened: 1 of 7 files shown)


tgoodlet:
I'll be moving this to a new repo (after extracting the git history) for further discussion and promotion!

PR #61, opened 2018-12-02
Faster row highlighting

tgoodlet:
This adds faster `piker monitor` *mouse over* row highlighting by using only one `on_mouse_pos()` callback and a loop over widgets in the highlight group.

In the worst case the loop has to iterate `N-1` widgets to find the next and last highlighted widget so I wonder if there's a more clever way to search for the next highlighted widget using a nearest neighbour heuristic or (if we wanted to generalize) a 2d metric to try and search the *closest* widgets to the last one.

ping @tshirtman!


piker/ui/kivy/mouse_over.py
```diff
@@ -0,0 +1,84 @@
...
+class MouseOverBehavior(object):
+    """Mouse over behavior.
+
+    :Events:
+        `on_enter`
+            Fired when mouse enter the bbox of the widget.
+        `on_leave`
+            Fired when the mouse exit the widget.
+    """
+    hovered = BooleanProperty(False)
```
tgoodlet: shoot, meant to change this to `mouse_over`

piker/ui/kivy/mouse_over.py
```diff
@@ -0,0 +1,84 @@
...
+    border_point = ObjectProperty(None)
+    _widgets = []
+    _last_hovered = None
+
+    def __init__(self, **kwargs):
+        self.register_event_type('on_enter')
+        self.register_event_type('on_leave')
+        MouseOverBehavior._widgets.append(self)
+        super(MouseOverBehavior, self).__init__(**kwargs)
+        Window.bind(mouse_pos=self.on_mouse_pos)
```
tshirtman: I assumed here this would only bind once, but if you have performance issues, i would check that assumption, as i'm not sure how bind plays with classmethod, maybe you have too many calls to `on_mouse_pos`. As mentionned, you could avoid it being called too often, by doing.
```
_on_mouse_pos = Clock.create_trigger(self.on_mouse_pos, timeout=0.01)
Window.bind(mouse_pos=_on_mouse_pos)
```

tgoodlet: Oh slick, I'll definitely try out the trigger.

tgoodlet:
@tshirtman also had some commentary on using a 2d binary search on discord:
>just got the point about binary search in 2D, if we scroll in both directions, it's a bit more of a pain, as you must know why it's not visible, so i'd binary search for the first visible with y < parent.top, and then probably do a binary search on the line end (y < first.y) and then a binary search for the first widget with right > parent.x, for each line (reusing the previously found line end for start of next binary search)
not sure how pleasant the code will be, and how much performance gained, but that probably depends on the shape of the grid, the ratio of visible/not visible cells, etc

Not sure if I'll get to that in this PR as it's more to do with generalizing this whole thing possibly for submission to the [garden](http://kivy-garden.github.io/).

piker/ui/kivy/mouse_over.py
tgoodlet: @tshirtman yeah so unfortunately this doesn't seem to work at all? The trigger doesn't call the `self.on_mouse_pos()` with the needed arguments - namely, the Window object and mouse coordinates.

Any suggestion to work around that? I looked at the [docs](https://kivy.org/doc/stable/api-kivy.clock.html#triggered-events) but don't really understand how triggers can pass through arguments to the underlying callback and it doesn't seem manually calling the event from a wrapper callback works at all?

PR #66, opened 2019-01-28
Tractor pub-sub

tgoodlet:
Port to the new pub-sub [`@tractor.msg.pub` system](https://github.com/tgoodlet/tractor/blob/master/tractor/msg.py#L90).


tgoodlet:
CI is still borked but that's coming in a separate PR.

PR #71, opened 2019-02-26
Fix ci

### LINKED ISSUE #6: CI for questrade broker backend
We can manually create a [*practice* server account](http://www.questrade.com/api/free-practice-account). Ideally we automate getting new practise credentials once expired (90 days).

I think first step should be using a single practise account and make sure the backend tests can run reliably.

tgoodlet:
Resolves  #6 as well as factors some CLI options to the top level `piker` command, namely:

- `-l` for log level
- `-c` for a configuration directory
- `-b` for the broker backend name

The QT practice account refresh token has to be set in the build env as `QT_REFRESH_TOKEN` and has to be deleted and reset if/when a new token has to be set. A [cached directory](https://docs.travis-ci.com/user/caching/#arbitrary-directories) is used to store a persistent brokers config (in the default `$HOME/.config/piker/brokers.ini` for now).

PS: the first time the build is run (before the cached broker config exists) the `travis` fixture seems to fail with the expected exception [not caught](https://github.com/pikers/piker/commit/414734f803d404f81fd57ea15abddbd3ac4e87e9#diff-484462fced51d1a06b1d93b4a44dd535R62)?. I'm pretty sure I reproduced this manually prior but I don't really have time to figure it out (or care) right now. It's probably something worth perfecting down the road.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,5 +16,9 @@ install:
     - cd $TRAVIS_BUILD_DIR
     - pipenv install --dev -e . --deploy
 
+cache:
+    directories:
+        - $HOME/.config/piker/
+
 script:
     - pipenv run pytest tests/
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -16,9 +16,9 @@
     "default": {
         "asks": {
             "hashes": [
-                "sha256:1679e5bd1dfa6c5d2220bdf2b8921c9c0d063d08370a7c66b9e167113681406f"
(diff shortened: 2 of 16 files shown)


tgoodlet:
Gahh whattt! The cache dir [isn't shared between builds!?!?](https://travis-ci.org/pikers/piker/builds/498527624#L190). Mannnn..so annoying.

tgoodlet:
Just turning off PR builds for now.
We'll stick with just push builds.

PR #112, opened 2020-09-02
Chart hacking

goodboy:
WIP first draft of rt charting.

Check the readme for testing [here](https://github.com/pikers/piker/tree/chart_hacking).

PR #113, opened 2020-09-11
Fixed some missing internal packages in setup install and added CI

guilledk:
This shows the error I encountered when trying to install `piker` on `ubuntu 18.04.5`, using `python 3.8.5`: https://github.com/guilledk/piker/actions/runs/250618249


setup.py
```diff
@@ -34,9 +34,11 @@
     platforms=['linux'],
     packages=[
         'piker',
+        'piker.cli',
+        'piker.data',
         'piker.brokers',
         'piker.ui',
-        'piker.testing',
+        'piker.testing'
```
goodboy, line 41: gonna need `piker.fsp` too... :crying_cat_face:

setup.py
```diff
@@ -34,9 +34,11 @@
     platforms=['linux'],
     packages=[
         'piker',
+        'piker.cli',
```
goodboy, line 37: Yeah, more I think about this it shouldn't be the cause of this import error since internally the packages use relative import syntax.

Not sure what's up..

goodboy: Ahh, I think it's that you're pip installing the master branch which is broken... If you were to make a `virtualenv` and install from this branch it would likely work.

goodboy: Also we should probably just use [`find_packages()`](https://setuptools.readthedocs.io/en/latest/setuptools.html#id8) for now.

goodboy:
@guilledk added this manually in 0556889 since history got desynced.

goodboy:
Replaced in #112 9b73952.
Thanks again for starting this @guilledk!

PR #117, opened 2020-11-06
Update README.rst

wattygetlood:


goodboy:
@wattygetlood we like the idea of making this more friendly to traders / retail but i would suggest a couple tweaks.
Would like to hear input from others as well.

README.rst
```diff
@@ -8,7 +8,7 @@ trading gear for hackers.
     :target: https://actions-badge.atrox.dev/piker/pikers/goto
 
 ``piker`` is a broker agnostic, next-gen FOSS toolset for real-time
-trading targeted at hardcore Linux users.
+trading targeted at retail traders (pikers) who are not satisfied with the market tools that are currently available.
```
goodboy, line 11: drop both *retail* and *(pikers)*

goodboy: maybe s/*with the market tools that are currently available.*/*current market tools*?

README.rst
```diff
@@ -47,12 +47,13 @@ focus and features:
 - geared for collaboration within trader communities
 - zero interest in adoption by suits; no corporate friendly license, ever.
 - not built for *sale*; built for *people*
+- tools designed, from the ground up, to allow traders to define their edge and execute it
```
goodboy, line 50: hmm, *edge* is kinda 80s-90s speak imo.
I wonder if something like:
*tools designed, from the ground up, to allow traders to express themselves*?

goodboy:
Yeah, gonna drop this in favor of a diff vibe which the crew has more favorably voted on 😂

PR #118, opened 2020-11-08
Add initial selection box interaction

goodboy:
I'd appreciate feedback from the low dpi peeps to be sure the `QLabel` use here is scaled right like our other font settings.

TODO:
- [ ] maybe add some more measures that seem good as defaults (eg. swing % high to low)?
- [ ] fix the `QLabel` bg color to match either the view's bg or something better matched to the selection box's hue?
- [ ] add support for multiple box annotations (maybe defer for another pr?)

feedback much appreciated.

Oh this also adds the start of keyboard controls which are horribly callback based but I'm not sure we want to fight the battle of trying to take input via `trio` just yet. Might be worth a peak at how far [qtrio](https://github.com/altendky/qtrio) has come on this front.

### DIFF
--- a/piker/ui/_interaction.py
+++ b/piker/ui/_interaction.py
@@ -142,6 +142,7 @@ class SelectRect(QtGui.QGraphicsRectItem):
 
         y1, y2 = start_pos.y(), end_pos.y()
         x1, x2 = start_pos.x(), end_pos.x()
+
         # TODO: heh, could probably use a max-min streamin algo here too
         ymn, xmn = min(y1, y2), min(x1, x2)
         ymx, xmx = max(y1, y2), max(x1, x2)
@@ -151,10 +152,15 @@ class SelectRect(QtGui.QGraphicsRectItem):
 
         ixmn, ixmx = round(xmn), round(xmx)
         nbars = ixmx - ixmn + 1
-        data = self._chart._array[ixmn:ixmx]
-        std = data['close'].std()
-        dmx = data['high'].max()
-        dmn = data['low'].min()
+
+        data = self._chart._ohlc[ixmn:ixmx]
+
+        if len(data):
+            std = data['close'].std()
+            dmx = data['high'].max()
+            dmn = data['low'].min()
+        else:
+            dmn = dmx = std = np.nan
 
         # update label info
         self._label.setText('\n'.join(self._contents).format(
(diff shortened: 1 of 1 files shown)


goodboy:
Lol probably should mention you gotta hold shift and then click drag to create a selection.

wattygetlood:
image
>100% move does not display % beside value

wattygetlood:
image
something is broken not seeing anything in the selection box when chart first starts up. XBTUSD^ kraken. It seems to fix itself after  time.

wattygetlood:
image
when selection box is drawn past the current time no data appears and chart lags significantly.

goodboy:
> image
> something is broken not seeing anything in the selection box when chart first starts up. XBTUSD^ kraken. It seems to fix itself after time.

Yeah this one I've seen as well; it's something to do with an initial bad state that shows up during startup.
The chart's *scene* layer has to load correctly before any of this works so there might be a race to test here.

goodboy:
> image
> when selection box is drawn past the current time no data appears and chart lags significantly.

@wattygetlood what do you expect to be shown in this case?
Just the number of bars / time maybe in the contents label?

wattygetlood:
> > image
> > when selection box is drawn past the current time no data appears and chart lags significantly.
> 
> @wattygetlood what do you expect to be shown in this case?
> Just the number of bars / time maybe in the contents label?
it should display in an identical manner wherever its located on the chart

goodboy:
@wattygetlood btw, on linux there's no extra lag in this case.

wattygetlood:
> @wattygetlood btw, on linux there's no extra lag in this case.

the lag may be gone once we have it properly functioning with IB and live data on windows

goodboy:
@wattygetlood so this is the issue list I have from you right now:

- [ ] metrics label doesn't include full text once numbers get too large (eg. 100% gainz)
- [ ] the weird `c` in the label (which I can't seem to reproduce any more so it might be fixed?)
- [ ] when the box is drawn outside the ohlc basis no contents are shown
  - this is a still problem **here on this branch** but will be fixed shortly by the new backing shared memory primary index adoption.
- [ ] lag, not seeing that at all now with the new graphics system we've adopted for OHLC (also coming soon).

I'm going to put these in a new issue (along with the initial list from above) where we can track making this better.

One the main things that now has my concern is that using `QWidget` *proxies* is apparently very slow already and a better solution might be to use `QGraphicsTextItem` for all of this.

PR #130, opened 2020-12-19
To qpainterpath and beyond

goodboy:
This re implements OHLC sampling graphics using the much faster [`QPainterPath`](https://doc.qt.io/qt-5/qpainterpath.html#composing-a-qpainterpath) which is how `pyqtgraph` gets decent speed for curves internally.

One of the nice properties of using this API instead is being able to append to existing paths very easily and they draw much faster for way larger data sets (think 3ms to draw 30-40k datums on screen). There's probably still further speedups to get with moving some internal routines in `pyqtgraph` to use `numba`. One of the major goals with all this is to get low latency interaction with a days worth of 1s sampled futures market data in view.

I won't go into too much detail because a lot of this was fairly involved tinkering and likely not super relevant to the average maintainer/contributor.

Some further fixes:
- kill the actor stack on window close gracefully
- fix some axis and font size stuff

New internal work:
- make shared mem arrays "start" indexing in the middle of the large backing array such that they can be backfilled with historical data loaded after the initial seed
- add support for prepending to shm segments which will auto-adjust the "start" index for readers of the consumable data
- adjust a bunch of internal UI indexing logic to compute correct "relative" indexes when accessing the "filled out" data segments from shm (probably need some better abstractions and APIs here)
- allow for brokers to provide a historical weighted average price field (wap) in the OHLC array if so defined in their module

### DIFF
--- a/piker/brokers/ib.py
+++ b/piker/brokers/ib.py
@@ -21,9 +21,10 @@ Note the client runs under an ``asyncio`` loop (since ``ib_insync`` is
 built on it) and thus actor aware API calls must be spawned with
 ``infected_aio==True``.
 """
-from contextlib import asynccontextmanager, contextmanager
+from contextlib import asynccontextmanager
 from dataclasses import asdict
 from functools import partial
+from datetime import datetime
 from typing import List, Dict, Any, Tuple, Optional, AsyncIterator, Callable
 import asyncio
 import logging
@@ -32,6 +33,7 @@ import itertools
 import time
 
 from async_generator import aclosing
+from ib_insync.wrapper import RequestError
 from ib_insync.contract import Contract, ContractDetails
 from ib_insync.ticker import Ticker
 import ib_insync as ibis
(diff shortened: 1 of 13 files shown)


piker/brokers/ib.py
```diff
@@ -346,9 +403,17 @@ async def _aio_get_client(
     """
     # first check cache for existing client
 
+    # breakpoint()
     try:
-        yield _client_cache[(host, port)]
-    except KeyError:
+        if port:
+            client = _client_cache[(host, port)]
```
goodboy, line 409: This actually wasn't working before and makes it so IB api clients are actually reused by multiple consuming actors; avoids client throttling that would result.

piker/brokers/ib.py
```diff
@@ -181,9 +209,13 @@ async def bars(
             # TODO: raise underlying error here
             raise ValueError(f"No bars retreived for {symbol}?")
 
+        # TODO: rewrite this faster with ``numba``
         # convert to pandas dataframe:
         df = ibis.util.df(bars)
-        return from_df(df)
+        return bars, from_df(df)
+
+    def onError(self, reqId, errorCode, errorString, contract) -> None:
```
goodboy, line 217: Heh, this i don't even think works.

PR #131, opened 2020-12-19
Vwap fsp

goodboy:
Initial [VWAP](https://www.tradingview.com/ideas/vwap/?solution=43000502018) fsp implementation using a standard "typical price" approach per bar.

This is still technically "tina" vwap in the sense that it isn't using historical tick data to compute the average yet. This is in theory possible though may be quite slow depending on broker historical query limitations (unless of course we make a way to collect the data offline). Either way I'm not sure how much the error from true vwap will actually be considering we're capable of using 1s OHLC with IB.

### DIFF
--- a/piker/brokers/ib.py
+++ b/piker/brokers/ib.py
@@ -192,6 +192,7 @@ class Client:
             # barSizeSetting='5 secs',
 
             durationStr='{count} S'.format(count=period_count),
+            # barSizeSetting='5 secs',
             barSizeSetting='1 secs',
 
             # barSizeSetting='1 min',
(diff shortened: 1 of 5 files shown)


piker/brokers/ib.py
```diff
@@ -328,8 +329,9 @@ async def find_contract(
             exch = 'SMART' if not exch else exch
             contract = (await self.ib.qualifyContractsAsync(con))[0]
 
-            head = await self.get_head_time(contract)
-            print(head)
+            # head = await self.get_head_time(contract)
```
goodboy, line 332: turns out we don't really need this unless looking up data to epoch - but even in that case pretty sure we'll just not receive datums past it.

PR #136, opened 2020-12-30
Add initial y-axis zoom

goodboy:
Y-axis zoom support.

Not totally convinced of the UX here yet, this is a pure manual control that can be reset by calling `ChartPlotWidget.default_view()` (which is triggered by the `r` keybinding currently).

What might make this better going forward is keeping our "max-min" range scaling based on the in-view data but only if the x-range changes such that data which is outside (greater then max or less then min) the current manually set y-range shows in view. This would mean you can only ever zoom the y "in" so far such that you don't clip max / min values in view. Not sure if this is what most peeps would want though so leaving it fully manual to start.

@wattygetlood

### DIFF
--- a/piker/ui/_chart.py
+++ b/piker/ui/_chart.py
@@ -384,7 +384,6 @@ class ChartPlotWidget(pg.PlotWidget):
 
         # self.setViewportMargins(0, 0, 0, 0)
         self._ohlc = array  # readonly view of ohlc data
-        self.default_view()
 
         self._arrays = {}  # readonly view of overlays
         self._graphics = {}  # registry of underlying graphics
@@ -405,6 +404,8 @@ class ChartPlotWidget(pg.PlotWidget):
 
         # show background grid
         self.showGrid(x=True, y=True, alpha=0.5)
+
+        self.default_view()
 
         # TODO: stick in config
         # use cross-hair for cursor?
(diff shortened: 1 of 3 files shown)


piker/ui/_chart.py
```diff
@@ -761,44 +770,46 @@ def _set_yrange(
...
-        diff = yhigh - ylow
-        ylow = ylow - (diff * 0.04)
-        yhigh = yhigh + (diff * 0.04)
+        if set_range:
+            # view margins: stay within a % of the "true range"
+            diff = yhigh - ylow
+            ylow = ylow - (diff * 0.04)
+            yhigh = yhigh + (diff * 0.04)
 
-        # # compute contents label "height" in view terms
```
goodboy, line 769: just moved all this old label size calc stuff into a new commented out method.

piker/ui/_interaction.py
```diff
@@ -295,6 +301,18 @@ def mouseDragEvent(self, ev, axis=None):
         # Scale or translate based on mouse button
         if ev.button() & (QtCore.Qt.LeftButton | QtCore.Qt.MidButton):
 
+            # zoom only y-axis when click-n-drag on it
```
goodboy, line 304: This is the core of it.
If a user click-drags on the y-axis we do scaling on the y-dimension only.

goodboy:
Though I still think this needs some work (see main comment above) there's probably no reason to hold off on this. It's kinda a must have as chart trading polishes up.

goodboy:
Yeah... thinking we need a mode and ability to pan y-dimension to make this more friendly - right now it's awkwardly keeping the y position pinned while zooming which will probably cause UX heck.

goodboy:
Gonna follow up in some kinda *view mode* PR.

PR #140, opened 2021-01-07
Basic alerts

goodboy:
Give this a shot boiyos (and/or girliyos and youiyos).

Controls:
- click (sub)plot on which to enter alert (probably need an indicator label for this)
- hold `a` key to stage line for alert
- *full click* location to submit alert "order"
- use `d` after mouse over of any line to delete it (still need to hook up deleting the actual alert

You'll need a notification daemon running, I use [`dunst`](https://dunst-project.org/documentation/).

It needs a ton more polish but I think the execution engine (aka EMS) is working well.

Let me know what y'all think!

ping @iamzoltan @guilledk

### DIFF
--- a/piker/_ems.py
+++ b/piker/_ems.py
@@ -1,0 +1,91 @@
+# piker: trading gear for hackers
+# Copyright (C) 2018-present Tyler Goodlet (in stewardship for piker0)
+
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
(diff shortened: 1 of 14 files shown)


iamzoltan:
Issue with the alert function. When placing the first alert, the window gets stuck in alert mode. You have to place a second alert and the press a once more to get out of alert mode. This seems like a bug. Once a is released, you should no longer be in alert mode.

goodboy:
@iamzoltan pretty sure it's fixed now in the basic orders stuff.

If you still have a problem open an issue and we'll dig in.

Issue #143, opened 2021-01-10
Supervise the store


goodboy:
It's about time we integrate `marketstor` as a "micro-daemon" spawned as part of [`pikerd`](https://github.com/pikers/piker/blob/master/piker/cli/__init__.py#L35) startup. More on `pikerd` re-design at end.

The current command I would expect to use is something like:
```
sudo docker run -v "/home/<user>/.config/piker/mkts.yml:/etc/mkts.yml" -v "/home/<user>/.config/piker/marketstore:/data" -p 5993:5993 alpacamarkets/marketstore:latest
```
which gets us
- latest released container
-  running with the data dir mapped to the `piker` user config dir
- with a `mks.yaml` config also in that same dir

The config we've been using is currently:
```yaml
# mount this config using:
# sudo docker run --mount type=bind,source="<path-to-piker>/piker/piker/data/configs",target="/etc" \
# -i -p 5993:5993 alpacamarkets/marketstore:latest
root_directory: data
listen_port: 5993
grpc_listen_port: 5995
log_level: info
queryable: true
stop_grace_period: 0
wal_rotate_interval: 5
stale_threshold: 5
enable_add: true
enable_remove: false

triggers:
  - module: ondiskagg.so
    on: "*/1Min/OHLCV"
    config:
        # filter: "nasdaq"
        destinations:
            - 1Min
            - 5Min
            - 15Min
            - 1H
            - 1D

  - module: stream.so
    on: '*/*/*'
    # config:
    #     filter: "nasdaq"
```

This will need to be updated as features from #142 land.

I propose the following a spawning and supervision:
- creat a supervising actor under `pikerd` and have it use `docker-py` to (pull and) run the container
- make some kind of sync mechanism to let other actors know the db can be used freely once up and running
- make this supervisor stream in logs from the container and and log them using piker logging where possible
- if there is a critical failure (container outright crashes) we should have come facility for respawn

It would further be great to get some of these container input params specced out in a piker config file.

In terms of a `pikerd` root service, we need to probably rejig this to use `tractor`'s new [implicit runtime stuff](https://github.com/goodboy/tractor/pull/185) as well as anticipate being the root process that launches UI components (Qt) in a child.

goodboy:
More explicit cmdline:
```
[...]

PR #149, opened 2021-01-31
Process tree structure refactor

guilledk:
This started trying to add the `marketstore` supervisor #143, but in the end it turned to a general refactor of the piker process tree.

### DIFF
--- a/piker/_daemon.py
+++ b/piker/_daemon.py
@@ -1,0 +1,79 @@
+"""
+pikerd daemon lifecylcle & rpc
+"""
+from typing import Optional
+from contextlib import asynccontextmanager
+
+import tractor
+
+from .log import get_logger, get_console_log
+from .brokers import get_brokermod
+
+
+log = get_logger(__name__)
+
+root_nursery: Optional[tractor._trionics.ActorNursery] = None
+root_dname = 'pikerd'
+root_modules = [
+    __name__,
+    'piker._ems'
+]
+
+@asynccontextmanager
+async def maybe_spawn_pikerd(
+    loglevel: Optional[str] = None
+) -> Optional[tractor._portal.Portal]:
+    """If no ``pikerd`` daemon-root-actor can be found,
+    assume that role and return a portal to myself
+    """
+    global root_nursery
+
+    if loglevel:
+        get_console_log(loglevel)
+
+    async with tractor.find_actor(root_dname) as portal:
+
+        if portal is not None:  # pikerd exists
+            yield portal
+
+        else:  # assume role
+            async with tractor.open_root_actor(
+                name=root_dname,
+                loglevel=loglevel,
+                enable_modules=root_modules
+            ) as actor:
+
+                # init root nursery
+                async with tractor.open_nursery() as nursery:
+                    root_nursery = nursery
+                    yield None
+
+
+# brokerd enable modules
+_data_mods = [
+    'piker.brokers.core',
+    'piker.brokers.data',
+    'piker.data',
+    'piker.data._buffer'
+]
+
+async def spawn_brokerd(
+    brokername,
+    loglevel: Optional[str] = None,
+    **tractor_kwargs
+):
+
+    brokermod = get_brokermod(brokername)
+    dname = f'brokerd.{brokername}'
+    log.info(f'Spawning {brokername} broker daemon')
+    tractor_kwargs = getattr(brokermod, '_spawnkwargs', {})
+
+    # TODO: raise exception when root_nursery == None?
+    global root_nursery
+    await root_nursery.start_actor(
+        dname,
+        enable_modules=_data_mods + [brokermod.__name__],
+        loglevel=loglevel,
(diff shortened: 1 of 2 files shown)


piker/_daemon.py
```diff
@@ -0,0 +1,79 @@
...
+    loglevel: Optional[str] = None,
+    **tractor_kwargs
+):
+
+    brokermod = get_brokermod(brokername)
+    dname = f'brokerd.{brokername}'
+    log.info(f'Spawning {brokername} broker daemon')
+    tractor_kwargs = getattr(brokermod, '_spawnkwargs', {})
+
+    # TODO: raise exception when root_nursery == None?
```
guilledk, line 71: if `root_nursery == None` here, calling this function doesn't make any sense, maybe we could do an `if` check and raise an exception but not sure which one would be apropiate.

piker/_daemon.py
```diff
@@ -0,0 +1,79 @@
...
+        else:  # assume role
+            async with tractor.open_root_actor(
+                name=root_dname,
+                loglevel=loglevel,
+                enable_modules=root_modules
+            ) as actor:
+
+                # init root nursery
+                async with tractor.open_nursery() as nursery:
+                    root_nursery = nursery
```
goodboy, line 48: maybe just put a comment here saying that this assigns the reference to the global module variable so that new calls to `pikerd` can access this same singleton nursery instance.

piker/_daemon.py
```diff
@@ -0,0 +1,79 @@
...
+    'piker.brokers.data',
+    'piker.data',
+    'piker.data._buffer'
+]
+
+async def spawn_brokerd(
+    brokername,
+    loglevel: Optional[str] = None,
+    **tractor_kwargs
+):
```
goodboy, line 64: Doc string would be good.

Just explain that this func is used to start sub-daemons under `pikerd` from remote client actors er wtv.

piker/_daemon.py
```diff
@@ -0,0 +1,79 @@
...
+
+    brokermod = get_brokermod(brokername)
+    dname = f'brokerd.{brokername}'
+    log.info(f'Spawning {brokername} broker daemon')
+    tractor_kwargs = getattr(brokermod, '_spawnkwargs', {})
+
+    # TODO: raise exception when root_nursery == None?
+    global root_nursery
+    await root_nursery.start_actor(
```
goodboy, line 73: Hmm another thing that might be handy here is caching the output portals to each name and then on repeat lookups you just noop if the daemon is already spawned?

piker/_daemon.py
```diff
@@ -0,0 +1,79 @@
+"""
```
goodboy, line 1: Add the license header.

We're being pedantic about this since the sw is likely to have a lot of attention in the future.

piker/_daemon.py
```diff
@@ -0,0 +1,79 @@
...
+
+root_nursery: Optional[tractor._trionics.ActorNursery] = None
+root_dname = 'pikerd'
+root_modules = [
+    __name__,
+    'piker._ems'
+]
+
+@asynccontextmanager
+async def maybe_spawn_pikerd(
```
goodboy, line 23: Hmm looking at this further I wonder if we should choose a better name like `maybe_open_pikerd`?

Technically this doesn't actually spawn anything, just might open a nursery.

piker/data/__init__.py
```diff
@@ -112,24 +97,25 @@ async def maybe_spawn_brokerd(
             yield portal
 
         else:  # no daemon has been spawned yet
+            async with maybe_spawn_pikerd(
```
goodboy, line 100: hmm or `maybe_start_pikerd()`?

piker/data/__init__.py
```diff
@@ -112,24 +97,25 @@ async def maybe_spawn_brokerd(
...
+                loglevel=loglevel
+            ) as pikerd_portal:
+
+                if pikerd_portal is None:
+                    await spawn_brokerd(
```
goodboy, line 105: Hmm this is interesting. I wonder if we can avoid having to check for `None` here (because we're likely going to see this pattern often)  and instead hand some kind of *local portal* thing kinda like [what's already inside `tractor`](https://github.com/goodboy/tractor/blob/master/tractor/_portal.py#L374) - but maybe it's over the top?

it would be nice to have the polymorphism to avoid all this branch logic.

goodboy (commented):
Code looks great!

I left a variety of questions, docs suggestions, and a bit of naming changes.
The one thing i wonder if we should consider is some kind of *local portal* shim to avoid all the branch logic for when we're "spawning" things locally. Not sure i'm convinced it's that big a deal yet tho.

goodboy:
@guilledk  I made a PR into your branch to update to the new clearing / order systems stuff.

I'm not sure if it's more of a hassle to get your branch updated and merge it or I'll just close this PR and make a new one with that updates branch?

goodboy:
Moved to #151.

PR #151, opened 2021-03-20
Supervise

goodboy:
Rebase and fixes to port to new clearing system of #149.

@guilledk I think this is probably the smoothest way forward to avoid waiting on you to rebase 😂 

We can both work off this branch now.


goodboy:
@guilledk so I think we can probably just land this since it's small enough and seems to mostly still work 😂 

the one last thing I was hoping to get going was running `pikerd` and being able to persist data feeds / shm arrays so that charts spawn faster when looking up an existing symbol and/or you've got alerts / orders that are still waiting to exec.

what you think?

piker/_daemon.py
```diff
@@ -0,0 +1,141 @@
+"""
```
goodboy, line 1: Ahh we need license header..

piker/_daemon.py
```diff
@@ -0,0 +1,141 @@
...
+# brokerd enabled modules
+_data_mods = [
+    'piker.brokers.core',
+    'piker.brokers.data',
+    'piker.data',
+    'piker.data._buffer'
+]
+
+
+async def spawn_brokerd(
```
goodboy, line 94: Do you think it's maybe just worth generalizing this spawn logic now?

For example we can have a `spawn_daemon(name: str)` and then just pass the module list in each case?

guilledk: Yeah that might be worth doing, easier to abstract shared behavoirs

goodboy:
Need to cherry-pick in a few commits from cached_feeds branch still.

Will probably delay the spawn api formalizing for another PR just to start keeping things small.

goodboy:
@guilledk pretty sure this is ready to land as well 🏄🏼

piker/_daemon.py
```diff
@@ -0,0 +1,228 @@
...
+
+_root_dname = 'pikerd'
+_root_modules = [
+    __name__,
+    'piker.clearing._ems',
+    'piker.clearing._client',
+]
+
+
+# @dataclass
```
goodboy, line 43: heh lingering comment.

piker/_daemon.py
goodboy: Kinda started this with the `_Services` type but we probably want to factor out the `spawn_<service>d()` code somewhat eventually as needed.

PR #153, opened 2021-03-28
Windows patches

wattygetlood:
Some minor fixes to get working on windows.

Might be more to come!

### DIFF
--- a/piker/data/_sharedmem.py
+++ b/piker/data/_sharedmem.py
@@ -22,7 +22,9 @@ from sys import byteorder
 from typing import List, Tuple, Optional
 from multiprocessing.shared_memory import SharedMemory, _USE_POSIX
 from multiprocessing import resource_tracker as mantracker
-from _posixshmem import shm_unlink
+
+if _USE_POSIX:
+    from _posixshmem import shm_unlink
 
 import tractor
 import numpy as np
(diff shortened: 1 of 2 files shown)


piker/ui/_exec.py
```diff
@@ -200,6 +200,7 @@ async def main():
                 tractor._root._default_arbiter_port,
             ),
             name='qtractor',
+            start_method='trio',
```
goodboy, line 203: I wish we had more peeps to test this 😂 

Would love to force this always.

goodboy: @wattygetlood actually, can you try without this after the history loading config changes i suggested?

piker/data/_sharedmem.py
```diff
@@ -73,7 +75,7 @@ def value(self) -> int:
 
     @value.setter
     def value(self, value) -> None:
-        self._shm.buf[:] = value.to_bytes(4, byteorder)
+        self._shm.buf[:] = value.to_bytes(self._shm.size, byteorder)
```
goodboy, line 78: Yah this is a good catch 🏄🏼

piker/ui/_exec.py
goodboy: Wtv we like `trio` more anyway 😂

goodboy (approved):
Lit.

PR #157, opened 2021-03-29
Hip shot: try making low dpi "just work"

goodboy:
There's a bunch more to do with this but figured I'd get it up so peeps can test.

So far most core devs and advanced users are working on hi dpi displays so HD has been mostly an afterthought.

This is a first draft/attempt at trying to make things "just work" when the auto-hidpi scaling doesn't seem to activate in minimal desktop linux envs (eg. i3 with no compositor).

This first bit of commits hacks font sizes depending on if a physical DPI of `<= 97` is detected; in which case a smaller set of default font sizes are used.

TODO:
- [x] scale a variety of annotation graphics for low dpi
- [x] scale pixel offsets for labels and axes based on some kind of font proportion?
- [ ] ~potentially get some users to test with full linux desktop envs (like gnome3) to see if the built-in hi-dpi detection stuff from Qt works out of the box~ deferring to be part of #159

### DIFF
--- a/piker/ui/_graphics/_cursor.py
+++ b/piker/ui/_graphics/_cursor.py
@@ -18,6 +18,7 @@
 Mouse interaction graphics
 
 """
+import math
 from typing import Optional, Tuple, Set, Dict
 
 import inspect
@@ -113,10 +114,6 @@ class LineDot(pg.CurvePoint):
         return False
 
 
-# TODO: likely will need to tweak this based on dpi...
-_y_margin = 5
-
-
 # TODO: change this into our own ``Label``
 class ContentsLabel(pg.LabelItem):
     """Label anchored to a ``ViewBox`` typically for displaying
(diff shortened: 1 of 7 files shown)


goodboy:
@guilledk give that latest tip a shot.

Pretty sure everything is resolved now 🏄🏼

piker/ui/_graphics/_lines.py
```diff
@@ -682,27 +682,27 @@ def order_line(
...
-            'order_type': order_type,
-            'level': level,
-            '$value': lambda f: f['level'] * f['size'],
-            'size': size,
-            'exec_type': exec_type,
-        }
-        llabel.orient_v = orient_v
-        llabel.render()
-        llabel.show()
+        # # left side label
```
goodboy, line 685: Yeah not convinced at all on these line labels minus the simple one for sizing. i'm thinkin we might want to swap out the contents labels with position and order info when order mode is engaged.

piker/ui/_graphics/_cursor.py
```diff
@@ -163,7 +170,9 @@ def __init__(
 
         ydim = margins[1]
         if inspect.isfunction(margins[1]):
-            margins = margins[0], ydim(font_size)
+            margins = margins[0], ydim(anchor_font_size)
+
+        print(f'margins: {margins}')
```
goodboy, line 175: drop this print

iamzoltan (commented):
Looks good to me. I like the idea of dpi aware fonts. What else are you looking to make dpi aware?

goodboy:
@iamzoltan pretty much everything in the UI/UX should just scale around information.

If you've used any decent software on mult-display gaming rigs you'll see that most good native apps just figure all this out for you and show a consistent "size in real world units" on screen - I think this is the way we should go from the outset.

iamzoltan:
yea, that sounds like the most sane approach

PR #158, opened 2021-03-31
Persistent data feeds

goodboy:
This is a rework on data feed machinery to allow for persistent real-time feeds to remain running after created by some data client where that client detaches but we want to keep the feed live for later use.

For now this means spawning a background feed task which also writes to shared memory buffers and can be accessed by any (returning) data client. This has the immediately useful benefit of allowing for super fast chart loading for pre-existing feeds.

Try it out by,
- in your first terminal running `pikerd`
- in your 2nd run `piker -b kraken XBTUSD`
- kill you chart
- re-run the chart command

Both the `ib` and `kraken` backends have been ported thus far.
Questrade can likely come in a new PR since the *single-connection-multi-symbol-subscription* style can also be accomplished with `kraken`'s websocket feed and we'll probably want to iron out details of how to request such things with both backends in view.

Some other stuff got slapped in here, namely engaging the Qt hidpi detection on windows (which reportedly works?)


piker/_daemon.py
```diff
@@ -87,6 +88,18 @@ async def open_pikerd(
             yield _services
 
 
+@asynccontextmanager
+async def maybe_open_runtime(
```
goodboy, line 92: Somewhat hacky and I wonder if we should offer this from `tractor` itself?

The trick is being able to ping the actor registry when you're code can run as a root out of process tree.

piker/_daemon.py
```diff
@@ -145,13 +160,28 @@ async def spawn_brokerd(
...
 
-    await _services.actor_n.start_actor(
+    portal = await _services.actor_n.start_actor(
         dname,
         enable_modules=_data_mods + [brokermod.__name__],
         loglevel=loglevel,
         **tractor_kwargs
     )
 
+    # TODO: so i think this is the perfect use case for supporting
```
goodboy, line 170: Link with goodboy/tractor#53, pretty sure the context entry / exit sync stuff would work amazing for this.

piker/data/_buffer.py
```diff
@@ -54,62 +60,92 @@ async def increment_ohlc_buffer(
...
+                    # this copies non-std fields (eg. vwap) from the last datum
+                    last[
+                        ['time', 'volume', 'open', 'high', 'low', 'close']
+                    ][0] = (t + delay_s, 0, close, close, close, close)
+
+                    # write to the buffer
+                    shm.push(last)
+
+                # broadcast the buffer index step
+                # yield {'index': shm._last.value}
```
goodboy, line 120: not sure why this comment is here.

piker/data/_buffer.py
```diff
@@ -54,62 +60,92 @@ async def increment_ohlc_buffer(
...
+                        await ctx.send_yield({'index': shm._last.value})
+                    except (
+                        trio.BrokenResourceError,
+                        trio.ClosedResourceError
+                    ):
+                        log.error(f'{ctx.chan.uid} dropped connection')
+
+
+@tractor.stream
+async def iter_ohlc_periods(
```
goodboy, line 132: This is effectively what the simplest dynamic (stream only exists while there is at least one subscriber) pub-sub system looks like.

I'm tempted to entirely drop the stuff in `msg.py` as the subscription API it self seems more confusing than this (`get_topics()`, `topics=List[str]`, and the like).

I think a thorough review and refactor with the `questrade` backend (which has a poll-with-multiple-symbols style producer system)  will probably illuminate what should be discarded.

piker/ui/_exec.py
```diff
@@ -87,16 +88,19 @@ def current_screen() -> QtGui.QScreen:
     assert screen, "Wow Qt is dumb as shit and has no screen..."
     return screen
 
-# XXX: pretty sure none of this shit works
+
+# XXX: pretty sure none of this shit works on linux as per:
```
goodboy, line 92: shoehorn alert.

goodboy:
Lol merged into wrong branch 😂

PR #161, opened 2021-04-06
Cached feeds

goodboy:
Copy of #158 but merging into `master` 😂 

This is a rework on data feed machinery to allow for persistent real-time feeds to remain running after created by some data client where that client detaches but we want to keep the feed live for later use.

For now this means spawning a background feed task which also writes to shared memory buffers and can be accessed by any (returning) data client. This has the immediately useful benefit of allowing for super fast chart loading for pre-existing feeds.

Try it out by,

    in your first terminal running `pikerd`
    in your 2nd run `piker -b kraken XBTUSD`
    kill you chart
    re-run the chart command and it should load faster and attach to the existing shm history

Both the `ib` and `kraken` backends have been ported thus far.

Questrade can likely come in a new PR since the single-connection-multi-symbol-subscription style can also be accomplished with kraken's websocket feed and we'll probably want to iron out details of how to request such things with both backends in view.

Some other stuff got slapped in here, namely engaging the Qt hidpi detection on windows (which reportedly works?)


goodboy:
Was actually good cause there was a conflict from the windows patches stuff.

piker/brokers/ib.py
```diff
@@ -811,78 +777,72 @@ def normalize(
...
-
-async def fill_bars(
+async def backfill_bars(
     sym: str,
-    first_bars: list,
-    shm: 'ShmArray',  # type: ignore # noqa
+    shm: ShmArray,  # type: ignore # noqa
     # count: int = 20,  # NOTE: any more and we'll overrun underlying buffer
-    count: int = 6,  # NOTE: any more and we'll overrun the underlying buffer
+    count: int = 10,  # NOTE: any more and we'll overrun the underlying buffer
```
goodboy, line 784: iirc this was < 6 to avoid tws throttling on another system I tried.

We might want to include this change as a default?

piker/data/_buffer.py
```diff
@@ -54,62 +60,92 @@ async def increment_ohlc_buffer(
...
+            await trio.sleep(ad)
+            total_s += lowest
+
+            # increment all subscribed shm arrays
+            # TODO: this in ``numba``
+            for delay_s, shms in _shms.items():
+                if total_s % delay_s != 0:
+                    continue
+
+                # TODO: numa this!
```
goodboy, line 98: Ahh right, that was the last thing we should do. Stick this in a new `_sampling.py` module instead of the current `data._buffer` which is a lame name.

piker/data/_sampling.py
```diff
@@ -0,0 +1,240 @@
...
+    subs.append(ctx)
+
+    try:
+        # stream and block until cancelled
+        await trio.sleep_forever()
+    finally:
+        subs.remove(ctx)
+
+
+async def sample_and_broadcast(
```
goodboy, line 154: There I think isolating these shm sampling routines is much more easy for code readers.

PR #167, opened 2021-04-15
Feed fixes

goodboy:
Fixes some minor issues integrating persistent feeds with the clearing engine as well as adds some polish for running deamons in `tractor`'s debug mode using a new `--pdb` flag.

Oh, and moved the `data/__init__.py` feed code into it's own submodule for easier organization in the future.

### DIFF
--- a/piker/clearing/_client.py
+++ b/piker/clearing/_client.py
@@ -137,8 +137,9 @@ def get_orders(
 
 
 # TODO: make this a ``tractor.msg.pub``
-async def send_order_cmds():
-    """Order streaming task: deliver orders transmitted from UI
+async def send_order_cmds(symbol_key: str):
+    """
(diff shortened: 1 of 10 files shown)


piker/clearing/_client.py
```diff
@@ -160,10 +161,18 @@ async def send_order_cmds():
     book._ready_to_receive.set()
 
     async for cmd in orders_stream:
-
-        # send msg over IPC / wire
-        log.info(f'Send order cmd:\n{pformat(cmd)}')
-        yield cmd
+        print(cmd)
```
goodboy, line 164: This is the major hack of note 😂 

We need proper broadcast channels as mentioned in https://github.com/goodboy/tractor/issues/204 to avoid this re-insertion blasphemy.

Issue #129, opened 2020-12-19
Selection box: upcoming features and bug fixes


goodboy:
Initial draft is going to land in #118 but there was a slew of follow bug reports and feature requests that we're going to list here and defer for now:

- [ ] maybe add some more measures that seem good as defaults (eg. swing % high to low)?
- [ ] fix the `QLabel` bg color to match either the view's bg or something better matched to the selection box's hue?
- [ ] add support for multiple box annotations (maybe defer for another pr?)
- [ ] metrics label doesn't include full text once numbers get too large (eg. 100% gainz) ![image](https://user-images.githubusercontent.com/291685/102697662-847d3580-4205-11eb-9f44-04c3d9067cf8.png)
- [ ] the weird `c` in the label (which I can't seem to reproduce any more so it might be fixed?) 
![image](https://user-images.githubusercontent.com/291685/102697858-2e10f680-4207-11eb-8511-8443ebf6cbba.png)

- [ ] when the box is drawn outside the ohlc basis no contents are shown 
![image](https://user-images.githubusercontent.com/291685/102697864-39fcb880-4207-11eb-8060-6c41cb54f9a3.png)

  - this has been fixed by the new backing shared memory primary index semantics.
- [ ] lag, not seeing that at all now with the new graphics system we've adopted for OHLC (also coming soon).

One the main things that now has my concern is that using `QWidget` *proxies* is apparently very slow already and a better solution might be to use `QGraphicsTextItem` for all of this.

PR #168, opened 2021-04-15
Symbol search

goodboy:
OK so i think this is finally ready for alpha testing!

#### Manual
- to open the search list: `ctl-space` or `ctrl-l`
  - the state from last open will be maintained
- to close the search list: `ctrl-space` (aka toggle) or `ctrl-c`
- once open, type pattern to match to provider symbols, they will appear in a per-provider results tree
- to navigate search results use either `up`/`down` arrow keys or `ctrl-j/k` (like in vim)
- to nav search result *provider-sections* you can use `ctrl-u/d` for up/down
- to select an entry (if it's not in the cache in which case it will switching automatically) use `enter`
- editting the search bar should be mostly as expected in default text editors (for now)

#### outstanding UX stuff todo (may get deferred to new issues / PRs):
- [x] no stupid blinking cursor
- [x] cache entries should be LIFO -> last selected should get moved to top and last recently before that should be next, etc
- [x] re-searching too often needs to be fixed; it results in more then necessary updates of the results tree
  - question: will this require updating specific segments of the results tree individually and how much more complicated will that make the `CompleterView.set_results()` implementation
  - providers that have already had a pattern request sent to search are searched again on slower provider debounce periods, ideally a patter search is idempotent and there should be a completion data structure that prevents *re-searches* once a provider has returned a results set for the input pattern.
  - UX feature: we could show a `searching...` entry for each provider that is currently not yet returned results
- [x] probably a `QLabel` in front of the `QLineEdit` for what kind of search we're doing 😂 
- [x] mouse selection support...not sure how i forgot this one
  
#### future wishlist
-> **made #186 to follow up on these**

```diff
- [ ] for `ib` unfortunately we need an ad-hoc futures / commodities / forex symbol set defined in the backend since they don't provide search for it; this needs to be added as a separate search task in ib's search routine
- [ ] character pattern highlighting in results entries
  - in `qutebrowser` [they re-invent the `QStyledItemDelegate`](https://github.com/qutebrowser/qutebrowser/blob/master/qutebrowser/completion/completiondelegate.py#L64) to accomplish this using `QTextEdit` 🤯 
  - here's a possible [dirty hack from the Qt forums](https://forum.qt.io/topic/61343/highlight-matched-substrings-in-qstyleditemdelegate/7)
- [ ] we need better sorting for some providers (eg. in `kraken` `adausd` will make a list with that exact match **not** at the top slot).
- [ ] better cross-provider sorting? right now i'm thinking provider order should be determined from a `provider.toml`?
- [ ] hotlists for personal positions across all asserts
- [ ] default hotlist from config (save this across chart sessions?)
- [ ] hotlist for all instruments with active clearing engine triggers/orders
- [ ] make `QLineEdit` cursor a rectangle with a nice contrasting color (like our bracket gray) 
```

### DIFF
--- a/piker/brokers/ib.py
+++ b/piker/brokers/ib.py
@@ -158,10 +158,19 @@ _adhoc_cmdty_data_map = {
 }
 
 _adhoc_futes_set = {
+
+    # equities
     'nq.globex',
     'mnq.globex',
     'es.globex',
     'mes.globex',
+
+    # cypto$
+    'brr.cmecrypto',
+    'ethusdrr.cmecrypto',
+
+    # metals
+    'xauusd.cmdty',
 }
 
 # exchanges we don't support at the moment due to not knowing
--- a/piker/brokers/ib.py
+++ b/piker/brokers/ib.py
@@ -833,6 +833,7 @@ async def get_bars(
 
     _err = None
 
+    fails = 0
     for _ in range(2):
         try:
 
@@ -847,7 +848,7 @@ async def get_bars(
 
             next_dt = bars[0].date
 
-            return bars, bars_array, next_dt
+            return (bars, bars_array, next_dt), fails
 
         except RequestError as err:
             _err = err
@@ -871,10 +872,13 @@ async def get_bars(
                     # and then somehow get that to trigger an event here
                     # that restarts/resumes this task?
                     await tractor.breakpoint()
+                    fails += 1
                     continue
 
-    else:  # throttle wasn't fixed so error out immediately
-        raise _err
+    return (None, None)
+
+    # else:  # throttle wasn't fixed so error out immediately
+    #     raise _err
 
 
 async def backfill_bars(
@@ -892,7 +896,7 @@ async def backfill_bars(
     https://github.com/pikers/piker/issues/128
 
     """
-    first_bars, bars_array, next_dt = await get_bars(sym)
+    (first_bars, bars_array, next_dt), fails = await get_bars(sym)
 
     # write historical data to buffer
     shm.push(bars_array)
@@ -904,9 +908,12 @@ async def backfill_bars(
         i = 0
         while i < count:
 
-            out = await get_bars(sym, end_dt=next_dt)
+            out, fails = await get_bars(sym, end_dt=next_dt)
 
-            if out is None:
+            if fails is None or fails > 1:
+                break
+
+            if out is (None, None):
                 # could be trying to retreive bars over weekend
                 # TODO: add logic here to handle tradable hours and only grab
                 # valid bars in the range
--- a/piker/brokers/ib.py
+++ b/piker/brokers/ib.py
@@ -179,6 +179,7 @@ _adhoc_futes_set = {
 _exch_skip_list = {
     'ASX',  # aussie stocks
     'MEXI',  # mexican stocks
+    'VALUE',  # no idea
 }
 
 # https://misc.interactivebrokers.com/cstools/contract_info/v3.10/index.php?action=Conid%20Info&wlId=IB&conid=69067924
@@ -832,7 +833,7 @@ async def get_bars(
 
     _err = None
 
-    for _ in range(1):
+    for _ in range(2):
         try:
 
             bars, bars_array = await _trio_run_client_method(
@@ -870,6 +871,7 @@ async def get_bars(
                     # and then somehow get that to trigger an event here
                     # that restarts/resumes this task?
                     await tractor.breakpoint()
+                    continue
 
     else:  # throttle wasn't fixed so error out immediately
         raise _err
@@ -1291,7 +1293,11 @@ async def open_symbol_search(
 
             if not pattern or pattern.isspace():
                 log.warning('empty pattern received, skipping..')
-                await stream.send(matches)
+
+                # XXX: this unblocks the far end search task which may
+                # hold up a multi-search nursery block
+                await stream.send({})
+
                 continue
 
             log.debug(f'searching for {pattern}')
--- a/piker/brokers/ib.py
+++ b/piker/brokers/ib.py
@@ -46,8 +46,8 @@ import ib_insync as ibis
 from ib_insync.wrapper import Wrapper
 from ib_insync.client import Client as ib_Client
 from fuzzywuzzy import process as fuzzy
+import numpy as np
 
-from .api import open_cached_client
 from ..log import get_logger, get_console_log
 from .._daemon import maybe_spawn_brokerd
 from ..data._source import from_df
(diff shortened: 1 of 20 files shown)


goodboy:
Might end up factoring out some commits from this one to help with getting backend support done sooner the later.

piker/brokers/ib.py
```diff
@@ -1179,3 +1267,63 @@ async def stream_trades(
             continue
 
         yield {'local_trades': (event_name, msg)}
+
+
+@tractor.context
+async def open_symbol_search(
```
goodboy, line 1273: uses `tractor`'s new context api to do request/response style searches over a stream.

piker/brokers/kraken.py
```diff
@@ -162,14 +179,51 @@ async def _public(
...
+
+        pairs = resp['result']
+
+        if pair is not None:
+            _, data = next(iter(pairs.items()))
+            return data
+        else:
+            return pairs
+
+    async def cache_symbols(
```
goodboy, line 202: we obviously need to change the name of this to something non-stock specific 😂

piker/brokers/kraken.py
```diff
@@ -162,14 +179,51 @@ async def _public(
...
+        self,
+    ) -> dict:
+        if not self._pairs:
+            self._pairs = await self.symbol_info()
+
+        return self._pairs
+
+    async def search_stocks(
```
goodboy, line 210: probably should just change this to `search_symbols()` 😂

piker/brokers/kraken.py
```diff
@@ -570,8 +638,34 @@ async def stream_quotes(
...
 
                 # XXX: format required by ``tractor.msg.pub``
                 # requires a ``Dict[topic: str, quote: dict]``
                 await send_chan.send({topic: quote})
+
+
+@tractor.context
+async def open_symbol_search(
+    ctx: tractor.Context,
+) -> Client:
```
goodboy, line 651: This return typing is probably wrong, not sure how we want that yet in `tractor`.

piker/data/_sampling.py
```diff
@@ -227,7 +227,8 @@ async def sample_and_broadcast(
             # end up triggering backpressure which which will
             # eventually block this producer end of the feed and
             # thus other consumers still attached.
-            subs = bus._subscribers[sym]
+            subs = bus._subscribers[sym.lower()]
```
goodboy, line 230: we're now internally normalizing all symbols to a lower case format for that *don't give af* style 😏

piker/ui/_event.py
```diff
@@ -0,0 +1,87 @@
+# piker: trading gear for hackers
```
goodboy, line 1: this is our event proxying from Qt subsystem.
basically, afaict this avoids 90% of the api *layers* used in `qtrio` but pretty much getting 90% of it's functionality -> getting Qt events relayed to async code over a mem chan.

piker/ui/_event.py
```diff
@@ -0,0 +1,87 @@
...
+
+        # never intercept the event
+        return False
+
+
+@asynccontextmanager
+async def open_key_stream(
+
+    source_widget: QtGui.QWidget,
+    event_type: QEvent = QEvent.KeyPress,
```
goodboy, line 72: Oh right this should be passed into the filter above...so that it's general smh.

piker/ui/_interaction.py
```diff
@@ -188,7 +188,7 @@ def set_pos(
 
         self._abs_top_right = label_anchor
         self._label_proxy.setPos(self.vb.mapFromView(label_anchor))
-        self._label.show()
+        # self._label.show()
```
goodboy, line 191: Ahh this is a me thing that leaked in.
Just trying to rethink the selection UX.

piker/ui/_search.py
```diff
@@ -0,0 +1,961 @@
...
+    _font,
+    DpiAwareFont,
+    # hcolor,
+)
+
+
+log = get_logger(__name__)
+
+
+class SimpleDelegate(QStyledItemDelegate):
```
goodboy, line 81: Can actually probably drop this for now but may end up being handy for syntax highlight.
See #186.

piker/ui/_search.py
```diff
@@ -0,0 +1,961 @@
...
+        # max_rows = 8  # 6 + search and headers
+        row_px = self.rowHeight(self.currentIndex())
+        # print(f'font_h: {font_h}\n px_height: {px_height}')
+
+        # TODO: probably make this more general / less hacky
+        self.setMinimumSize(self.width(), rows * row_px)
+        self.setMaximumSize(self.width() + 10, rows * row_px)
+        self.setFixedWidth(333)
+
+    def is_selecting_d1(self) -> bool:
```
goodboy, line 189: can remove this?

piker/ui/_search.py
```diff
@@ -0,0 +1,961 @@
...
+
+            ix = QStandardItem(str(i))
+            item = QStandardItem(s)
+
+            # Add the item to the model
+            section_item.appendRow([ix, item])
+
+        self.expandAll()
+
+        # TODO: figure out if we can avoid this line in a better way
```
goodboy, line 398: This bit here running on every section results insert is what's causing the selection to jump around when re-searches take place while typing. it's super annoying and basically we want to just select whatever the *last* selection was if possible to avoid this.

piker/ui/_search.py
```diff
@@ -0,0 +1,961 @@
...
+
+                elif key in {Qt.Key_J, Qt.Key_Down}:
+                    item = view.select_next()
+
+                if item:
+                    parent_item = item.parent()
+
+                    if parent_item and parent_item.text() == 'cache':
+
+                        value = search.get_current_item()
```
goodboy, line 906: can probably change this to use `SearchWidget.chart_selected_item()` too?

piker/ui/_interaction.py
goodboy: #129 related.

piker/ui/_search.py
goodboy: Nah, gonna leave it in, can always trim later.

piker/ui/_search.py
goodboy: link to #186

piker/ui/_search.py
goodboy: made another comment on how we could try figuring out the "abstract row,colum, parent" position once the user starts navigating and even if entries are swapped out underneath we just jump to the same "relative selection position" when possible.

Again can be worked out in #186.

piker/ui/_search.py
goodboy: Done.

piker/ui/_event.py
goodboy: done.

piker/ui/_event.py
goodboy: done.

piker/brokers/kraken.py
goodboy: done, plus got all backends verified working with it using `piker search <symbol>` 🥳

Issue #135, opened 2020-12-30
Move to using `httpx`


goodboy:
Apparently [`asks`](https://github.com/theelous3/asks) isn't being maintained (as much?) any more [and that's what we're using](https://github.com/pikers/piker/search?q=asks). All the cool kids saying `httpx`?

The `httpx` port [looks pretty easy though](https://www.python-httpx.org/async/).

PR #170, opened 2021-05-07
Add Binance backend

guilledk:
Binance seems to have a pretty acceptable API.

[General Info](https://binance-docs.github.io/apidocs/spot/en/#general-info)
[REST OHLC for historical backfilling](https://binance-docs.github.io/apidocs/spot/en/#kline-candlestick-data)
[Websocket L1 stream](https://binance-docs.github.io/apidocs/spot/en/#individual-symbol-ticker-streams)
[Realtime trade feed](https://binance-docs.github.io/apidocs/spot/en/#trade-streams)

### DIFF
--- a/piker/brokers/binance.py
+++ b/piker/brokers/binance.py
@@ -63,14 +63,17 @@ _ohlc_dtype = [
     ('low', float),
     ('close', float),
     ('volume', float),
-    # XXX: don't need these in shm history right?
+    ('bar_wap', float),  # will be zeroed by sampler if not filled
+
+    # XXX: some additional fields are defined in the docs:
+    # https://binance-docs.github.io/apidocs/spot/en/#kline-candlestick-data
+
     # ('close_time', int),
     # ('quote_vol', float),
     # ('num_trades', int),
     # ('buy_base_vol', float),
     # ('buy_quote_vol', float),
     # ('ignore', float),
-    ('bar_wap', float),  # will be zeroed by sampler if not filled
 ]
 
 # UI components allow this to be declared such that additional
@@ -106,33 +109,12 @@ class Pair(BaseModel):
     permissions: List[str]
 
 
-# TODO: this isn't being used yet right?
 @dataclass
 class OHLC:
     """Description of the flattened OHLC quote format.
 
     For schema details see:
     https://binance-docs.github.io/apidocs/spot/en/#kline-candlestick-streams
-
-    documented format:
-    ```
-        [
-          [
-            1499040000000,      // Open time
-            "0.01634790",       // Open
-            "0.80000000",       // High
-            "0.01575800",       // Low
-            "0.01577100",       // Close
-            "148976.11427815",  // Volume
-            1499644799999,      // Close time
-            "2434.19055334",    // Quote asset volume
-            308,                // Number of trades
-            "1756.87402397",    // Taker buy base asset volume
-            "28.46694368",      // Taker buy quote asset volume
-            "17928899.62484339" // Ignore.
-          ]
-        ]
-    ```
 
     """
     time: int
@@ -154,9 +136,6 @@ class OHLC:
     # null the place holder for `bar_wap` until we
     # figure out what to extract for this.
     bar_wap: float = 0.0
-
-    # (sampled) generated tick data
-    # ticks: List[Any] = field(default_factory=list)
 
 
 # convert arrow timestamp to unixtime in miliseconds
@@ -263,17 +242,17 @@ async def get_client() -> Client:
 
 # validation type
 class AggTrade(BaseModel):
-    e: str  # "aggTrade",  # Event type
-    E: int  # 123456789,   # Event time
-    s: str  # "BNBBTC",    # Symbol
-    a: int  # 12345,       # Aggregate trade ID
-    p: float  # "0.001",     # Price
-    q: float  # "100",       # Quantity
-    f: int  # 100,         # First trade ID
-    l: int  # 105,         # Last trade ID
-    T: int  # 123456785,   # Trade time
-    m: bool  # true,        # Is the buyer the market maker?
-    M: bool  # true         # Ignore
+    e: str  # Event type
+    E: int  # Event time
+    s: str  # Symbol
+    a: int  # Aggregate trade ID
+    p: float  # Price
+    q: float  # Quantity
+    f: int  # First trade ID
+    l: int  # Last trade ID
+    T: int  # Trade time
+    m: bool  # Is the buyer the market maker?
+    M: bool  # Ignore
 
 
 async def stream_messages(ws):
--- a/piker/brokers/binance.py
+++ b/piker/brokers/binance.py
@@ -55,28 +55,6 @@ log = get_logger(__name__)
 _url = 'https://api.binance.com'
 
 
-# Broker specific ohlc schema (websocket)
-_websocket_ohlc_dtype = [
-    ('index', int),
-    ('time', int),
-    ('close_time', int),
-    ('symbol', str),
-    ('interval', str),
-    ('first_trade_id', int),
-    ('last_trade_id', int),
-    ('open', float),
-    ('close', float),
-    ('high', float),
-    ('low', float),
-    ('volume', float),
-    ('num_trades', int),
-    ('closed', bool),
-    ('quote_asset_volume', float),
-    ('taker_buy_base_asset_volume', float),
-    ('taker_buy_quote_asset_volume', float),
-    ('ignore', int)
-]
-
 # Broker specific ohlc schema (rest)
 _ohlc_dtype = [
     ('index', int),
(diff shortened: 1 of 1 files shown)


piker/brokers/binance.py
```diff
@@ -0,0 +1,425 @@
...
+        'method': 'SUBSCRIBE',
+        'params': [
+            f'{pair.lower()}@{sub_name}'
+            for pair in pairs
+        ],
+        'id': uid
+    }
+
+
+class AutoReconWs:
```
guilledk, line 258: Maybe put this in some sort of `utils` module or something like that, as it is shared between kraken & binance backends.

goodboy: Yah 💯.

We need to fix up the kraken ws respawning anyway so we should probably stick that generic subs mgmt code somewhere as well.

goodboy:
@guilledk lol so as mentioned in chat, the quote rates coming from binance easily fall in the "higher" frequency realm 😂 

```
binance quote rate (Hz): 201.71711633722887
binance quote rate (Hz): 176.93752372917106
binance quote rate (Hz): 552.900606380174
binance quote rate (Hz): 168.3918419784808
binance quote rate (Hz): 110.53058212770443
binance quote rate (Hz): 491.8850709510965
binance quote rate (Hz): 579.7241188666206
binance quote rate (Hz): 48.043618702893404
binance quote rate (Hz): 13.001965343005052
binance quote rate (Hz): 134.26928740636404
binance quote rate (Hz): 25.902277555456745
binance quote rate (Hz): 1047.2669163545568
binance quote rate (Hz): 46.08973330549542
binance quote rate (Hz): 14.019004836440086
binance quote rate (Hz): 255.75024390243902
binance quote rate (Hz): 207.49500346294647
binance quote rate (Hz): 30.968214472936157
binance quote rate (Hz): 810.3369397217929
binance quote rate (Hz): 172.3639352346511
binance quote rate (Hz): 111.79146565740025
binance quote rate (Hz): 690.4204115226338
binance quote rate (Hz): 279.41536206781694
binance quote rate (Hz): 66.22724688940819
binance quote rate (Hz): 1018.529383195726
binance quote rate (Hz): 686.1285784393915
binance quote rate (Hz): 240.80284762888965
binance quote rate (Hz): 1101.7346992382454
```

In which case we should be throttling graphics consumers to their local display / refresh rate to avoid unnecessary render cycles.

Pretty happy with the look of things so far 🏄🏼

piker/brokers/binance.py
```diff
@@ -64,12 +63,14 @@
     ('low', float),
     ('close', float),
     ('volume', float),
-    ('close_time', int),
-    ('quote_vol', float),
-    ('num_trades', int),
-    ('buy_base_vol', float),
-    ('buy_quote_vol', float),
-    ('ignore', float)
+    # XXX: don't need these in shm history right?
```
goodboy, line 66: @guilledk I was gonna ask, I don't think we need this for now in the historical data right?

I don't really understand a couple of the fields actually.
Maybe we should make an issue to dig in later?

piker/brokers/binance.py
```diff
@@ -99,42 +101,67 @@ class Pair(BaseModel):
...
+# TODO: this isn't being used yet right?
 @dataclass
 class OHLC:
     """Description of the flattened OHLC quote format.
 
     For schema details see:
-        https://binance-docs.github.io/apidocs/spot/en/#kline-candlestick-streams
+    https://binance-docs.github.io/apidocs/spot/en/#kline-candlestick-streams
+
+    documented format:
```
goodboy, line 117: If you think we can drop the example (bc self explanatory from API docs) I'm cool with that too.

piker/brokers/binance.py
```diff
@@ -198,12 +228,29 @@ async def bars(
...
+            for j, (name, ftype) in enumerate(_ohlc_dtype[1:]):
+
+                # TODO: maybe we should go nanoseconds on all
+                # history time stamps?
+                if name == 'time':
+                    # convert to epoch seconds: float
+                    row.append(bar.time / 1000.0)
+
+                else:
+                    row.append(getattr(bar, name))
```
goodboy, line 251: This felt really really clunky.

What's a better way to one shot validate these?
I don't know if you can unpack `*bar` in a `pydantic` mode..

piker/brokers/binance.py
```diff
@@ -214,15 +261,32 @@ async def get_client() -> Client:
     yield Client()
 
 
+# validation type
+class AggTrade(BaseModel):
```
goodboy, line 265: The example fields can probably be cleaned out here?

piker/brokers/binance.py
```diff
@@ -239,6 +303,25 @@ async def stream_messages(ws):
                 ]
             }
 
+        elif msg.get('e') == 'aggTrade':
+
+            # validate
+            msg = AggTrade(**msg)
+
+            # TODO: type out and require this quote format
```
goodboy, line 311: This is definitely part of writing up the backend spec.

piker/brokers/binance.py
guilledk: Yeah I just put them for completeness but most are not necessary, I'll just leave the comment

piker/brokers/binance.py
guilledk: Cool

piker/brokers/binance.py
```diff
@@ -0,0 +1,496 @@
...
+    # ('quote_vol', float),
+    # ('num_trades', int),
+    # ('buy_base_vol', float),
+    # ('buy_quote_vol', float),
+    # ('ignore', float),
+]
+
+# UI components allow this to be declared such that additional
+# (historical) fields can be exposed.
+ohlc_dtype = np.dtype(_ohlc_dtype)
```
goodboy, line 81: Hmm it's interesting i think in this case we don't actually need to define this since [it matches the default](https://github.com/pikers/piker/blob/master/piker/data/_source.py#L29) but, i guess for now documenting that we could have extras is handy?

piker/brokers/binance.py
```diff
@@ -0,0 +1,496 @@
...
+
+# convert arrow timestamp to unixtime in miliseconds
+def binance_timestamp(when):
+    return int((when.timestamp() * 1000) + (when.microsecond / 1000))
+
+
+class Client:
+
+    def __init__(self) -> None:
+        self._sesh = asks.Session(connections=4)
```
goodboy, line 149: Oh that was the other thing.

What do you think about swapping out [`httpx`](https://www.python-httpx.org/async/) here?

Apparently everyone in `trio` is saying it's the more maintained project?

goodboy: We can defer to #135 as well if you want.

piker/brokers/binance.py
```diff
@@ -0,0 +1,496 @@
...
+
+            bar = OHLC(*bar)
+
+            row = []
+            for j, (name, ftype) in enumerate(_ohlc_dtype[1:]):
+
+                # TODO: maybe we should go nanoseconds on all
+                # history time stamps?
+                if name == 'time':
+                    # convert to epoch seconds: float
```
goodboy, line 226: We should make an issue about possibly moving to nanoseconds as `int` for all time stamps.
I see no reason to not be thinking about UHFT down the road.

piker/brokers/binance.py
goodboy: Yeah I mucked with this for a while and got nowhere too fancy 😂 

Since our dtype is technically matching the dataclass we could actually probably just use `asdict` if we wanted to do it this way:

```python
[nav] In [116]: @dataclass
           ...: class C:
           ...:     doggy: int
           ...:     num: int

[nav] In [117]: l = [C(doggy=9, num=10), C(doggy=11, num=12)]

[nav] In [118]: np.array(list(list(asdict(c).values()) for c in l), dtype={'names': ('doggy', 'num',), 'formats': (int, int)})
Out[118]:
array([[( 9,  9), (10, 10)],
       [(11, 11), (12, 12)]], dtype=[('doggy', '<i8'), ('num', '<i8')])
```

goodboy: Using a `pydantic.BaseModel.dict()` would require array unpacking each `bar` into dict form since I still don't think `pydantic.BaseModel` supports the `OHLC(*bar)` style creation.

Not sure it'd be worth it anyway for readability.

piker/brokers/binance.py
```diff
@@ -0,0 +1,496 @@
...
+    while True:
+
+        with trio.move_on_after(5):
+            msg = await ws.recv_msg()
+
+        # for l1 streams binance doesn't add an event type field so
+        # identify those messages by matching keys
+        # https://binance-docs.github.io/apidocs/spot/en/#individual-symbol-book-ticker-streams
+
+        if msg.get('u'):
```
goodboy, line 268: Oh that was my other question: can we do similar `BaseModel` unpacking/validation here for `l1` messages?

goodboy:
@guilledk oh one more thing would be to find out if binance supports historical tick data [like kraken](https://support.kraken.com/hc/en-us/articles/218198197-How-to-retrieve-historical-time-and-sales-trading-history-using-the-REST-API-Trades-endpoint-) such that we can get long(er) term views of high frequency sampled data (1s OHLC) going. If not we may have to look to some other historical resources.

Making at least issue for this task would be good.

PR #172, opened 2021-05-16
Minor fixes to make piker installable again

guilledk:
Pinned some dependencies, and updated an `arrow` timestamp API used in `kraken` backend.


goodboy (approved):
Nice catch.

Ye we gotta update in dev envs more often especially as onlookers come on board.

PR #174, opened 2021-05-17
TectonicDB support

guilledk:


piker/data/tectonicdb.py
```diff
@@ -0,0 +1,154 @@
...
+
+from typing import Union
+
+import trio
+import struct
+
+import piker.data.ffi
+
+
+class TectonicDB():
```
goodboy, line 25: You don't need the `()`.

Issue #126, opened 2020-12-04
Py3.9 support, numba bork-fest.


goodboy:
We're currently borked due to numba/numba#6345.

One heads up on arch is you need `llvm10` pkg.

goodboy:
https://github.com/numba/numba/issues/6345#issuecomment-775873609 seems to say it's working now so we can probably move forward with 3.9 finally 🥳

PR #176, opened 2021-05-20
Py3.9

goodboy:
Try pinning to python 3.9 and run what tests we can from the original suite.


goodboy:
TODO:
- [ ] pin to python 3.9 going foward in `setup.py`?
- [ ] close out #126 

maybe more?

goodboy:
Aight test suite is workin now that all the questrade stuff is commented out.

Gonna pin to 3.9 then I think this is ready to land.

@guilledk @iamzoltan py3.9 is coo by y'all yah?

PR #177, opened 2021-05-22
Naively throttle graphics updates to 60 FPS

goodboy:
For user testing.

This helps reduce cpu cycles on less beefy machines with high frequency feeds like from `binance`.
There's really no point in updating faster then the refresh rate anyway for manual viewing purposes.

TODO:
Deferred to #183.

Some further resources re: v-sync (though not sure if we should dig this deep yet):
- [SO question](https://stackoverflow.com/a/28457021) on syncing images with refresh rate
  - answer points to [SO on `updateGL()` realtime in Qt](https://stackoverflow.com/questions/17167194/how-to-make-updategl-realtime-in-qt) which requires V-sync
- [SO on missing frames with v-sync](https://stackoverflow.com/questions/37239900/qt-vsync-missing-rendered-frames) and a matching [Qt forum post](https://forum.qt.io/topic/67103/vsync-missing-rendered-frames)

### DIFF
--- a/piker/ui/_chart.py
+++ b/piker/ui/_chart.py
@@ -18,6 +18,7 @@
 High level Qt chart widgets.
 
 """
+import time
 from typing import Tuple, Dict, Any, Optional, Callable
 from types import ModuleType
 from functools import partial
@@ -934,7 +935,9 @@ async def test_bed(
         # txt.setPos(vb_right - w, d_coords.y())
         # txt.setPlainText(f"FUCK YOU {i}")
         i += 1
-        # rlabel.setPos(vb_right - 2*w, d_coords.y())
+
+
+_quote_throttle_rate = 60  # Hz
 
 
 async def chart_from_quotes(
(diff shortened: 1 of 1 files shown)


goodboy:
@guilledk let's get this base in and I'll make an issue about improving it.

binance is a bit laggy for me on a less beefy system without this 😂

goodboy:
Made and issue so this should be good to go 🏄🏼

piker/ui/_chart.py
```diff
@@ -1011,9 +1014,17 @@ def maxmin():
     tick_size = chart._lc.symbol.tick_size
     tick_margin = 2 * tick_size
 
+    last = time.time()
     async for quotes in stream:
-        for sym, quote in quotes.items():
 
+        now = time.time()
+        period = now - last
+        if period <= 1/_quote_throttle_rate - 0.001:
```
guilledk, line 1022: Maybe use `timeit` here

https://stackoverflow.com/a/25823885

goodboy: Toss that in #183 if you want.

As discussed in chat, i don't think the faults of `time.time()` are so relevant at display refresh rates 😂 

Also the top answer on that SO does suggest `time.time()`.

PR #182, opened 2021-05-24
Binance updates

### LINKED ISSUE #180: `binance` todos
Thanks once again to @guilledk for the new `binance` backend support in #170 we can get some more serious (and near hft) style processing going on for crypto$ 😎 

Some lingering things we need to get going stat as follow ups to that PR:
- [ ] we can probably clean up the manual loop `OHLC` bar-rows construction we're doing using [the suggestion i commented](https://github.com/pikers/piker/pull/170#discussion_r637424423) 
- [ ] we might not [actually need to define `ohlc_dtype`](https://github.com/pikers/piker/pull/170#discussion_r637419651) in the backend module (other then for docs of historical datum fields) since right now we aren't storing any more then the [minimal required field set](https://github.com/pikers/piker/blob/master/piker/data/_source.py#L29)
- [ ] we can [probably do `pydantic.BaseModel` validation](https://github.com/pikers/piker/pull/170#discussion_r637425128) on `l1` feed messages

---

New features we need for the backend to go prime time, these can come incrementally:
- [ ] longer out historical data retreival which has a couple options:
  - the [old trades](https://binance-docs.github.io/apidocs/spot/en/#old-trade-lookup) which appears to be tick data we can aggregate (1k datums per request and requires `X-MBX-APIKEY` in the request)
  - the [historical OHLC ep](https://binance-docs.github.io/apidocs/spot/en/#kline-candlestick-data) which *i think* is limited to 1m sampling period? (would be great if they did smaller periods for us).
- [ ] symbol search support using the symbol set delivered from the [exchange info endpoint](https://binance-docs.github.io/apidocs/spot/en/#exchange-information) -> this will likely land as part of #164 
- [ ] order entry via order mode / chart trader by providing EMS support
  - this needs some follow up from `piker`'s end in terms of docs and a standard messaging spec
  - relevant [spot order endpoints section](https://binance-docs.github.io/apidocs/spot/en/#spot-account-trade) looks great and includes a [order test endpoint](https://binance-docs.github.io/apidocs/spot/en/#spot-account-trade)
  - position and order tracking can be done [with one off requests](https://binance-docs.github.io/apidocs/spot/en/#account-information-user_data) or using a [real time updated stream](https://binance-docs.github.io/apidocs/spot/en/#payload-account-update) -> we'll definitely want support for the latter in our `emsd` subsys.

guilledk:
Relates to #180 

Since there's more then one part.

### DIFF
--- a/piker/brokers/binance.py
+++ b/piker/brokers/binance.py
@@ -45,6 +45,8 @@ log = get_logger(__name__)
 
 _url = 'https://api.binance.com'
 
+# XXX: some additional fields are defined in the docs:
+# https://binance-docs.github.io/apidocs/spot/en/#kline-candlestick-data
 ohlc_dtype = np.dtype(ohlc_with_index)
 
 
@@ -233,13 +235,11 @@ class Client:
 
         new_bars = []
         for i, bar in enumerate(bars):
-
-            init_dict = {}
-            for j, key in enumerate(
-                KLine.__dict__['__fields__'].keys()):
-                init_dict[key] = bar[j]
-
-            bar = KLine(**init_dict)
+            bar = KLine(**{
+                key: bar[j]
+                for j, key in enumerate(
+                    KLine.__dict__['__fields__'].keys())
+            })
             bar_wap = .0
             new_bars.append((i,) + tuple(bar.as_row()) + (bar_wap,))
 
--- a/piker/brokers/binance.py
+++ b/piker/brokers/binance.py
@@ -457,6 +457,7 @@ class Client:
         async with trio.open_nursery() as n:
             n.start_soon(periodic_keep_alive, key)
             yield key
+            n.cancel_scope.cancel()
 
         await self.close_listen_key(key)
 
@@ -731,10 +732,9 @@ async def handle_order_requests(
                 )
 
             elif action == 'cancel':
-                # msg = BrokerdCancel(**request_msg)
-                # 
-                # await client.submit_cancel(symbol, msg.reqid)
-                ...
+                msg = BrokerdCancel(**request_msg) 
+
+                await client.submit_cancel(msg.symbol, msg.reqid)
 
             else:
                 log.error(f'Unknown order command: {request_msg}')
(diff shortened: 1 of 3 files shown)


piker/brokers/binance.py
```diff
@@ -47,38 +47,16 @@
...
-    ('time', int),
-    ('open', float),
-    ('high', float),
-    ('low', float),
-    ('close', float),
-    ('volume', float),
-    ('bar_wap', float),  # will be zeroed by sampler if not filled
-
-    # XXX: some additional fields are defined in the docs:
-    # https://binance-docs.github.io/apidocs/spot/en/#kline-candlestick-data
```
goodboy, line 69: I think we should still leave the link to the api.

piker/brokers/binance.py
```diff
@@ -207,31 +194,23 @@ async def bars(
...
-            bar = OHLC(*bar)
+            init_dict = {}
+            for j, key in enumerate(
+                KLine.__dict__['__fields__'].keys()):
+                init_dict[key] = bar[j]
 
-            row = []
-            for j, (name, ftype) in enumerate(_ohlc_dtype[1:]):
+            bar = KLine(**init_dict)
+            bar_wap = .0
```
goodboy, line 206: just inline this.

piker/brokers/binance.py
```diff
@@ -240,7 +219,16 @@ async def get_client() -> Client:
     yield Client()
 
 
-# validation type
+# validation types
+class BookTicker(BaseModel):
```
goodboy, line 223: nice!

goodboy:
@guilledk you should consider adding `-b binance` to our [example chart cmd](https://github.com/pikers/piker#check-out-our-charts) too and maybe add a little section on how to search for multiple symbols?

Either way imo binance is probably the more prime cex that peeps gonna want to see out of the box.

goodboy:
Oh also, this needs to be rebased don't forget.

piker/brokers/binance.py
```diff
@@ -141,16 +159,56 @@ def __init__(self) -> None:
...
-        resp = await self._sesh.get(
-            path=f'/api/v3/{method}',
-            params=params,
-            timeout=float('inf')
-        )
+
+        if signed:
+            params['signature'] = self._get_signature(params)
+
+        if action == 'get':
```
guilledk, line 197: @goodboy Here is a part I'm not confident in, you think I should just make separate methods for `get` / `post` or this is fine?

piker/brokers/binance.py
```diff
@@ -245,33 +303,76 @@ async def bars(
...
+        recv_window: int = 60000
+    ) -> int:
+        symbol = symbol.upper()
+
+        await self.cache_symbols()
+
+        asset_precision = self._pairs[symbol]['baseAssetPrecision']
+        quote_precision = self._pairs[symbol]['quoteAssetPrecision']
+
+        quantity = Decimal(quantity).quantize(
```
guilledk, line 342: We need to sanitize `price`/`quantity` because `binance` has specific per-symbol requirements on precision.

guilledk:
Still `TODO`:

- Order status updates
- Order cancel

piker/brokers/binance.py
goodboy: Yeah this is also going to play in with new $positioning features where a user can specify the base pair for the position size; which fits naturally with crypto / forex, less so with discrete instruments like stocks / most derivatives.

piker/brokers/binance.py
```diff
@@ -141,16 +159,56 @@ def __init__(self) -> None:
...
 
+        conf = get_config()
+        self.api_key = conf.get('api', {}).get('key')
+        self.api_secret = conf.get('api', {}).get('secret')
+
+        if self.api_key:
+            self._sesh.headers.update({'X-MBX-APIKEY': self.api_key})
+
+    def _get_signature(self, data: OrderedDict) -> str:
+        if not self.api_secret:
```
guilledk, line 170: Should probably add some more comments liking to docs about the signing of requests

piker/brokers/binance.py
goodboy: Hmm good question.
This seems fine I think?

you could maybe even just proxy the method direct to `asks` -> `getattr(self._sesh, method)` ?

Not sure if that's too "dynamic" or wtv.

guilledk: Yeah love that option, always forget about `getattr`

piker/brokers/binance.py
```diff
@@ -493,6 +685,135 @@ async def subscribe(ws: wsproto.WSConnection):
...
+            f'wss://stream.binance.com:9443/ws/{listen_key}',
+        ) as ws:
+            event = await ws.recv_msg()
+
+            if event.get('e') == 'executionReport':
+                """
+                https://binance-docs.github.io/apidocs/spot/en/#payload-balance-update
+                """
+
+                oid = event.get('c')
```
guilledk, line 763: I'll write a pydantic model for all this later

goodboy:
@guilledk this needs a rebase eh?

guilledk:
Yeah, I'm catching up with $dayjob, so will not be able to continue this for a bit, this should be almost done pretty sure.

goodboy:
Replaced by the lurker sponsored #510 💥 / 🏄🏼

Issue #164, opened 2021-04-08
Symbol search


goodboy:
Oh boy, well this will probably be a somewhat drawn out / involved UI/UX endevour but i'm gonna start by tossing in a bunch of resources, most getting thrown to me for free 🏄🏼 

- the `Qt` [custom completer example](https://doc.qt.io/qt-5/qtwidgets-tools-customcompleter-example.html)
- [`QSortFilterProxyModel`](https://doc.qt.io/qt-5/qsortfilterproxymodel.html)
- youtube's [getting started with `PyQt5` playlist](https://www.youtube.com/playlist?list=PL3JVwFmb_BnRpvOeIh_To4YSiebiggyXS)
- lda.org's [how to implement autocomplete](https://learndataanalysis.org/how-to-implement-autocomplete-feature/)
- lda's [python pyqt5 tutorial's list](https://learndataanalysis.org/category/python-tutorial/python-pyqt5/)
- some sweet auto-magic packaging via [`copier-poetry-fbs`](https://gitlab.com/mikeramsey/copier-poetry-fbs)
- [`copier`](https://copier.readthedocs.io/en/stable/) itself
-

goodboy:
Fuzzy search stuff:
- https://github.com/taleinat/fuzzysearch#examples
- https://news.ycombinator.com/item?id=20772683

PR #184, opened 2021-05-25
Binance syminfo and mintick

goodboy:
Some stuff we missed in `binance` backend.

- correct formatting output of `.symbol_info()`
- symbol set cacheing at client startup (in prep for #164 feature requirements)
- extract min tick info per symbol to allow charting axes to resolve correctly.

@guilledk you should rebase your #182 off this 😎

### DIFF
--- a/piker/brokers/binance.py
+++ b/piker/brokers/binance.py
@@ -288,10 +288,20 @@ class AggTrade(BaseModel):
 
 
 async def stream_messages(ws):
+
+    timeouts = 0
     while True:
 
-        with trio.move_on_after(5):
+        with trio.move_on_after(5) as cs:
             msg = await ws.recv_msg()
+
+        if cs.cancelled_caught:
+
+            timeouts += 1
+            if timeouts > 2:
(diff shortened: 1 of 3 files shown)


piker/brokers/binance.py
```diff
@@ -492,5 +545,32 @@ async def stream_quotes(
...
+    async with open_cached_client('binance') as client:
+
+        # load all symbols locally for fast search
+        cache = await client.cache_symbols()
+        await ctx.started()
+
+        async with ctx.open_stream() as stream:
+
+            async for pattern in stream:
+                # results = await client.symbol_info(sym=pattern.upper())
```
goodboy, line 565: Yeah turns out they don't have a "search" in the API they offer, you have to have the exact symbol.

Collecting the whole cache at client startup is reasonably fast anyway (and we do it async with the other backends), so all gud i think.

guilledk (approved):

goodboy:
Ahh gotta push up a `tractor` patch to expose `Portal` in order to fix CI.

PR #188, opened 2021-05-27
Web utils: no more reconnect bs

goodboy:
Add a re-subscription *fixture* to the "reliable" websocket API such that subs are rerun on reconnects.
That stuff now sits in it's own module factored out of the crypto backends that require it.

Also added a slight tweak to the graphics rate limiting such that l1 udpates are limited to twice a slow as the refresh rate since there's really no need for it on the visual feed.

Testing this now but so far seems like it's working as intended 😎

### DIFF
--- a/piker/brokers/binance.py
+++ b/piker/brokers/binance.py
@@ -18,36 +18,26 @@
 Binance backend
 
 """
-from contextlib import asynccontextmanager, AsyncExitStack
-from types import ModuleType
+from contextlib import asynccontextmanager
 from typing import List, Dict, Any, Tuple, Union, Optional
-import json
 import time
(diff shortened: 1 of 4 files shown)


guilledk (approved):

piker/brokers/binance.py
```diff
@@ -311,14 +300,15 @@ async def stream_messages(ws):
     timeouts = 0
     while True:
 
-        with trio.move_on_after(5) as cs:
+        with trio.move_on_after(3) as cs:
```
goodboy, line 303: @guilledk just want to make sure there was no reason this had to be a longer wait 🤔

PR #189, opened 2021-05-31
Basic status bar

goodboy:
Initial simple status bar to show startup sequence and current UI "mode" in a right side label.
This branch will require a `pip install -U qdarkstyle` to work since I ported to latest version.

More to come for this but it's probably ready to land as a simple starting point.

Future todos which might get moved to a new issue:
- [x] add support for status *tokens* which let async code set a status, receive a token obj that can then be called to clear that status later
- [x] support multiple (concurrenct) statuses via a `{status_1, status_2, status_3}` type text display for multiple tasks?
- ~[ ] override the default splitter background color (pretty sure it should be our `default_dark` alias)~ leaving this for now, the color grew on me.
- ~[ ] per provider/broker feed/ems statuses which down the road can be augmented with some widgets for resetting or displaying info?~ made #197 
  - thinking we should defer this to a new issue that can come with the throttling stuff in #177 ?

ping @iamzoltan @guilledk

### DIFF
--- a/piker/ui/_chart.py
+++ b/piker/ui/_chart.py
@@ -508,6 +508,8 @@ class ChartPlotWidget(pg.PlotWidget):
     def focus(self) -> None:
         # self.setFocus()
         self._vb.setFocus()
+        app = QtGui.QApplication.instance()
+        app.activeWindow().statusBar().showMessage("mode: view")
 
     def last_bar_in_view(self) -> int:
         self._ohlc[-1]['index']
@@ -1661,6 +1663,8 @@ async def _async_main(
 
     chart_app = widgets['main']
 
+
+
     # attempt to configure DPI aware font size
     screen = current_screen()
 
@@ -1674,6 +1678,10 @@ async def _async_main(
 
     # configure global DPI aware font size
     _font.configure_to_dpi(screen)
+
+    sbar = chart_app.window.statusBar()
+    sbar.setStyleSheet(f"font: {_font.px_size - 3}px")
+    sbar.showMessage('starting ze chartz...')
 
     async with trio.open_nursery() as root_n:
 
@@ -1700,6 +1708,7 @@ async def _async_main(
         symbol, _, provider = sym.rpartition('.')
 
         # this internally starts a ``chart_symbol()`` task above
+        sbar.showMessage(f'loading {provider}.{symbol}...')
         chart_app.load_symbol(provider, symbol, loglevel)
 
         root_n.start_soon(load_providers, brokernames, loglevel)
--- a/piker/ui/_chart.py
+++ b/piker/ui/_chart.py
@@ -52,7 +52,6 @@ from ._l1 import L1Labels
 from ._graphics._ohlc import BarItems
 from ._graphics._curve import FastAppendCurve
 from ._style import (
-    _config_fonts_to_screen,
     hcolor,
     CHART_MARGINS,
     _xaxis_at,
@@ -68,7 +67,7 @@ from ..data import maybe_open_shm_array
 from .. import brokers
 from .. import data
 from ..log import get_logger
-from ._exec import run_qtractor, current_screen
+from ._exec import run_qtractor
 from ._interaction import ChartView
 from .order_mode import start_order_mode
 from .. import fsp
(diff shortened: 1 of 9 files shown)


piker/ui/_exec.py
```diff
@@ -124,6 +149,40 @@ def closeEvent(
...
+        # style and cached the status bar on first access
+        if not self._status_bar:
+            # TODO: i guess refactor stuff to avoid having to import here?
+            from ._style import _font, hcolor
+            sb = self.statusBar()
+            sb.setStyleSheet((
+                f"color : {hcolor('gunmetal')};"
+                f"background : {hcolor('default_dark')};"
+                # "min-height : 19px;"
+                f"font-size : {_font.px_size - 4}px;"
```
guilledk, line 164: Bar itself looks great, on low dpi monitors tho we have a font size issue, on my PC the correct font size setting would be: `font-size : {_font.px_size + 1}px;`

guilledk (changes requested):

piker/ui/_exec.py
goodboy: kk, might need a small adhoc condition then to adjust when dpi is detected as 96ish.

piker/ui/_chart.py
```diff
@@ -1673,7 +1698,13 @@ async def _async_main(
     log.info(f'Set graphics update rate to {_clear_throttle_rate} Hz')
 
     # configure global DPI aware font size
-    _font.configure_to_dpi(screen)
+    _config_fonts_to_screen()
+
+    # TODO: do styling / themeing setup
```
goodboy, line 1703: drop this

piker/ui/_exec.py
```diff
@@ -114,6 +145,32 @@ def __init__(self, parent=None):
...
+        self._status_bar: QStatusBar = None
+        self._status_label: QLabel = None
+
+    @property
+    def mode_label(self) -> QtGui.QLabel:
+
+        # init mode label
+        if not self._status_label:
+            # TODO: i guess refactor stuff to avoid having to import here?
+            from ._style import _font_small, hcolor
```
goodboy, line 157: Hmm i wonder if it's worth it to just now move all this stuff to a new `._window.py` mod?

I'm thinking it'll be a nice cleanup.

piker/ui/_search.py
```diff
@@ -765,16 +770,22 @@ async def fill_results(
...
+                    # input pattern (in which case just look up the old
+                    # results).
                     if (period >= pause) and (
                         provider not in already_has_results
                     ):
+
+                        # TODO: it may make more sense TO NOT search the
+                        # cache in a bg task since we know it's fully
+                        # cpu-bound.
+                        if provider != 'cache':
```
goodboy, line 785: This drastically improved re-searching lag btw.

piker/ui/_style.py
```diff
@@ -22,18 +22,22 @@
...
 log = get_logger(__name__)
 
+_magic_inches = 0.0666 * (1 + 6/16)
+
 # chart-wide fonts specified in inches
 _font_sizes: Dict[str, Dict[str, float]] = {
     'hi': {
-        'default':  0.0616,
-        'small':  0.055,
+        # 'default':  0.0616,
```
goodboy, line 37: might be able to drop these old values.

piker/ui/_style.py
```diff
@@ -95,22 +100,34 @@ def configure_to_dpi(self, screen: Optional[QtGui.QScreen] = None):
...
+        if mx_dpi <= 97:  # for low dpi use larger font sizes
             inches = _font_sizes['lo'][self._font_size]
-        else:
+
+        else:  # hidpi use smaller font sizes
             inches = _font_sizes['hi'][self._font_size]
 
+        dpi = mn_dpi
+
+        # dpi is likely somewhat scaled down so use slightly larger font size
```
goodboy, line 115: would love to see if this works on windows 😉

piker/ui/_exec.py
```diff
@@ -124,6 +181,40 @@ def closeEvent(
...
+    @property
+    def status_bar(self) -> QStatusBar:
+
+        # style and cached the status bar on first access
+        if not self._status_bar:
+            # TODO: i guess refactor stuff to avoid having to import here?
+            from ._style import _font_small, hcolor
+            sb = self.statusBar()
+            sb.setStyleSheet((
+                f"color : {hcolor('gunmetal')};"
```
goodboy, line 193: btw i **really** like this font color for widget text; nice and muted but still matches.

piker/ui/_exec.py
goodboy: Did this.

Issue #192, opened 2021-06-02
HF feeds and UI performance via throttling


goodboy:
Since the addition of `binance` provider support in #170 i've noticed some serious cpu usage on weaker hw due the high data rates received on the binance websocket pipes. The L1 quote rates in particular can be instantaneous rates 500-1kHz which far and above now only most 60FPS display rates but even higher end stuff.

This issue can be considered a partial followup subtask for #180.

#### Toolz
- probably worth checking out [`perf-timer`'s async support](https://github.com/belm0/perf-timer#async-support) to figure out exactly which tasks are causing the most cpu gobble

---
#### Premises
- for `Qt` UIs we shouldn't need to update graphics more then 60 FPS from a data feed, any more then this is just wasted cycles
- the high rate feeds don't seem to pose any issues with processing latency/cpu usage so there should be no need to throttle for fsp, clearing or algorithmic purposes
- currently inside chart code we've [naively throttled](https://github.com/pikers/piker/blob/master/piker/ui/_chart.py#L1002) the quote and book rates but, [we're still task switching and ignoring](https://github.com/pikers/piker/blob/master/piker/ui/_chart.py#L1102) too fast arrival rates; **this is likely causing `Qt` task switches at greater then refresh rates which, for data viz purposes, is likely pointless**.
- rough measures show that CPU usage starts really climbing with ` > 5` symbols loaded in the local chart cache; you'll start to notice the usage in a sys monitor as well as light latency when scrolling/panning in the view box with the mouse.
  - this all depends on ur hw of course 😂 
- we're currently running a [websocket-per-symbol-subscription](https://github.com/pikers/piker/blob/0da02aa2601dc39e5e1da73fb216ca2f789d3ac9/piker/brokers/binance.py#L469) meaning there's a separate ws for each loaded chart
  - in theory, more ws connections may mean more context switching overhead in `trio`, though it seems this would be doubting the OS's IO latency?
  - more ws connections i don't *think* helps resiliency much since if the actor thread dies we lose all the conns versus just the ws dying we lose a symbol, but there's no reason we can't just have multiple ws with all the subs going? It's worth some thinking.

---
#### Solution ideas to try
- [ ] get a better serialization backend in `tractor`?
  - tried out `msgspec` in https://github.com/goodboy/tractor/pull/212 but the improvement is negligible (as would be expected for encode only atm)
- [ ] move throttling to the `data.feed` layer?
[...]

Issue #109, opened 2020-08-24
Chart graphics with drawing speedups


goodboy:
Had a great chat with a collaborator discussing some options for squeezing performance out of `pyqtgraph`.
A lot was confirming ideas I had previous but a few nuggets are going to be particularly fun to experiment with.
Also note, there is no real latency problem yet but this is anticipation of updating multiple assets per plot. 

This will all be more relevant once the first draft of #10 lands :surfing_man: (soooon).

As part of that landing will come a naive charting update system using some well known slowly-ness like `np.append()`/`np.concatenate()` (which frankly aren't anywhere near being a bottleneck yet) but which we can of course improve upon.

The tip(s) received for improving rendering latency and general datum compacting include:
- charts are resized much less frequently than interactions within them - allocate a `numpy` array 4x the size, and use `numba` to do the *subsetting* (aka downsampling in graphics land)
- the general strategy for said *downsampling* (if drawing tick-like, event triggered lines) is something like:
  > only 4 points per pixel
you bin by the width of the chart, and put the entry val, min, max, exit in that bin
then it will be as accurate looking as the full data
  - this idea comes from 2 papers related to a data viz downsampling algorithm:
    - [M4: A Visualization-Oriented Time Series Data Aggregation](http://www.vldb.org/2014/program/papers/demo/p997-jugel.pdf)
    - [Faster Visual Analytics through Pixel-Perfect Aggregation](http://www.vldb.org/pvldb/vol7/p797-jugel.pdf)
      - >look at fig 3 for why you need 4 points
otherwise you get discontinuities as you zoom
the idea is to have the rendering look the same for the sampled version as if you threw all the data in the range (without having the chart library try to do something smart, just render all data using a standard line drawing algo)
anyway that's only if you care about pixel width blips ;)
- `pyqtgraph` has some similar implementations for lines in the code base that could be potentially improved with this:
  - the [`clipToView` algo code](https://github.com/pyqtgraph/pyqtgraph/blob/a719a06b8b8e6c642aeffd2f6797124fa5a2cd82/pyqtgraph/graphicsItems/PlotDataItem.py#L581)
  - the [`downSampleMethod == 'peak'` code](https://github.com/pyqtgraph/pyqtgraph/blob/a719a06b8b8e6c642aeffd2f6797124fa5a2cd82/pyqtgraph/graphicsItems/PlotDataItem.py#L615)
[...]

Issue #107, opened 2020-08-20
Numpy compatible shared mem ring buffer


goodboy:
As part of the [fsp](https://en.wikipedia.org/wiki/Financial_signal_processing) subsystem design we're likely going to want to implement some shared memory systems for low latency multi-time frame calcs. I've started a couple design ideas about this in #98 but I do think for sure part of this system will require a fast ring buffer for `numpy` data:

Here's a starter list of projects to try after some very brief searching

##### committed to long ago
- 2016: [ringbuffer](https://github.com/bslatkin/ringbuffer) it's multi-reader/writer, multi-proc aware and has an [example with `numpy`](https://github.com/bslatkin/ringbuffer/blob/master/example_numpy.py#L46)
- 2015: [`pyringbuf`](https://github.com/sirlark/pyringbuf) a c-ext implementing a *character buffer*

##### committed recently
- 2020: [`pyring`](https://github.com/jaycosaur/pyring) pure python impl and has a [multi-proc c-types example](https://github.com/jaycosaur/pyring#multiprocess-c-types-example)
- 2020: [`ringbuf`](https://github.com/elijahr/ringbuf) looks to be built for real-time audio but is single producer/consumer (which is actually fine for our use case afaik) but is built as a cython wrapper around `boost` (ala C++) libs; `numpy` [example looks good](https://github.com/elijahr/ringbuf#numpy)
- 2020: [circular_buffer_numpy](https://github.com/vstadnytskyi/circular_buffer_numpy) seems kinda like a simple wrapper around an array with "pointer" index references kinda like what i've rolled many times before.

It may just end up we take a lang from #106 and implement one or try out some designs from above on the new [`SharedMemory`](https://docs.python.org/3/library/multiprocessing.shared_memory.html#multiprocessing.shared_memory.SharedMemory) type which has an example using a `numpy` array.

As always, lurkers please chime.

goodboy:
[data_pipe](https://github.com/random-python/data_pipe) is another one to check out (though not sure it'll have `numpy` support at all.

Looks to imply it has some [disruptor style](https://github.com/random-python/data_pipe#usage) examples.

goodboy:
This definitely can tie in with #192 @guilledk

PR #199, opened 2021-06-15
Naive feed throttling

goodboy:
First attempt at addressing #192 in a very very naive way.

Summary:
- port `data.feed.open_feed()` machinery to use `@tractor.context` apis
- add a `data._sampling.uniform_rate_send()` quote limiter which allows for queuing L1 ticks and throttling quote transmission to clients (like the charts) at a constant rate (like 60 FPS).

There's still lots of much fancier stuff that can be done from here as mentioned in #107 for fast shm tick queuing and in #109 for potential graphics side optimizations.

I think this is at least providing a good start point for further queuingn eng work 😎

### DIFF
--- a/piker/data/_sampling.py
+++ b/piker/data/_sampling.py
@@ -17,6 +17,7 @@
 """
 Data buffers for fast shared humpy.
 """
+import time
 from typing import Dict, List
 
 import tractor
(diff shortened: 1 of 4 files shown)


iamzoltan (approved):
looks good!

guilledk (approved):

piker/data/_sampling.py
```diff
@@ -229,16 +231,71 @@ async def sample_and_broadcast(
...
-                    trio.ClosedResourceError
-                ):
-                    # XXX: do we need to deregister here
-                    # if it's done in the fee bus code?
-                    # so far seems like no since this should all
-                    # be single-threaded.
-                    log.error(f'{ctx.chan.uid} dropped connection')
+            for (stream, tick_throttle) in subs:
+
+                if tick_throttle:
```
goodboy, line 236: I feel like we could just make the mem chan look like the `MsgStream` here and avoid the flag check, but not gonna stress about it right now.

PR #200, opened 2021-06-16
Asyncify input modes

goodboy:
This converts our `ChartView` kb handling code to **very succinct** async loop (`qtloops` anyone?) and basically makes all the callback-based state tracking moot since it can be massively simplified using edge triggered key detection and `set` operations 🏄🏼 

This also fixes a bunch of longstanding UX bugs to do with `live` vs. `dark` order mode being active depending on mod keys.

Additions:
- extended the `Qt` event relaying api with `trio` to include a nice `open_handler()` for registering an async func to handle events from a user specified type set.
- re-orged a bunch of code that was in `ui._interaction.py` into new modules (namely `_editors.py`).
- added the `s` key for *submit* to activate "live order mode" vs. `ctrl` (which you can still use for now.

Further kb control additions:
- `cc` to cancel all orders on a chart
- `q` is now required to display "historical quotes", i.e. OHLC values won't be displayed in a contents label by default without edge triggering this key.


piker/ui/_chart.py
```diff
@@ -77,13 +73,23 @@
 log = get_logger(__name__)
 
 
-class ChartSpace(QtGui.QWidget):
-    '''Highest level composed widget which contains layouts for
+class GodWidget(QtGui.QWidget):
+    '''
+    "Our lord and savior, the holy child of window-shua, there is no
+    widget above thee." - 6|6
+
```
guilledk, line 80: Praise be!

guilledk (approved):

piker/ui/_chart.py
goodboy: 🙏🏼 📈

PR #202, opened 2021-07-05
Minimal `brokerd` trade dialogues

goodboy:
Reworks `emsd` to keep a minimal set of `brokerd` `trades_dialogue()` endpoint streams using caching and an actor-local map to determine the client side stream for which to relay trade events.

 More pragmatically this avoids having multiple of the same event routed to clients who are not the originators of the trade dialogue on client-ems side while keeping a single stream (where possible) on the `brokerd` side. More generally, this also allows us to move toward client-leg filtering of trade flows as well as eventual trade dialogue mirroring and tracking.
 
 ~Oh this also includes a small rework of the daemon APIs to avoid use of the `AsyncExitStack` for allocating remote `tractor.Context` connections which is pretty non-ideal since any errors that propagate from daemons won't be captured until the stack is unwound - which then kind of defeats the point of it in the first place. Instead we just allocate tasks via the `pikerd` root nursery which wait on both the context and the surrounding portal such that all errors will propagate upward.~
 
 ^ Was pulled out and landed in #203.


goodboy (commented):
There pulled out the daemon stuff into #203 making that stuff distinct and much simpler from this more monumental patch set.

piker/clearing/_ems.py
```diff
@@ -80,15 +84,16 @@ def check_lt(price: float) -> bool:
...
-    Contains conditions for executions (aka "orders") which are not
-    exposed to brokers and thus the market; i.e. these are privacy
-    focussed "client side" orders.
+    Contains conditions for executions (aka "orders" or "triggers")
+    which are not exposed to brokers and thus the market; i.e. these are
+    privacy focussed "client side" orders which are submitted in real-time
+    based on specified trigger conditions.
 
-    A singleton instance is created per EMS actor (for now).
+    A an instance per `brokerd` is created per EMS actor (for now).
```
guilledk, line 94: Small typo, should be "An instance per..." right?

guilledk (approved):
Very readable as always!

goodboy:
CI for the install being weird but 🤷🏼

PR #203, opened 2021-07-06
Wait on daemon portals, concurrently.

goodboy:
Rework our daemon "service manager" to drop the use of `AsyncExitStack` and instead use plain tasks, spawned in the `pikerd` actor, to wait on sub-daemons and bubble up any errors captured therein.

The original use of an async exit stack is very problematic since we have no way to collect errors from opened `tractor.Context`'s *to* sub-daemons *as these errors arrive*. This is of course because even if you wait on the exit of the stack the LIFO ordering may not capture an error from an earlier *pushed* daemon portal (and thus propagate it to the surrounding nursery / cancel scope) since these contexts are **not** being waited upon **concurrently**.

Instead spawn a task for each sub-daemon which both waits on the opened context and portal. Now when any daemon errors, `pikerd` will see that immediately and deal with it accordingly (which is right now just to bail as well).

### DIFF
--- a/piker/_daemon.py
+++ b/piker/_daemon.py
@@ -19,11 +19,12 @@ Structured, daemon tree service management.
 
 """
 from typing import Optional, Union, Callable, Any
-from contextlib import asynccontextmanager, AsyncExitStack
+from contextlib import asynccontextmanager
 from collections import defaultdict
 
 from pydantic import BaseModel
 import trio
+from trio_typing import TaskStatus
 import tractor
 
 from .log import get_logger, get_console_log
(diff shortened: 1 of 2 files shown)


guilledk (approved):
Very clean refactor!

piker/_daemon.py
```diff
@@ -117,22 +161,22 @@ async def open_pikerd(
...
         ) as _,
+
         tractor.open_nursery() as actor_nursery,
     ):
         async with trio.open_nursery() as service_nursery:
 
-            # setup service mngr singleton instance
-            async with AsyncExitStack() as stack:
+            # # setup service mngr singleton instance
+            # async with AsyncExitStack() as stack:
```
goodboy, line 170: lel need to clean this.

PR #204, opened 2021-07-09
Ib adhoc derivs search

goodboy:
Start building out a derivatives and commodities symbol map inside the `ib` backend since there is no search for such symbols in the API (only equities).

todo:
- add a search routine over the ad-hoc symbols dict
- i'd like to add in support for loading existing orders on api connect
-  for orders that get "delayed" (like say no stocks to borrow for shorting) sometime the order gets submitted to the *when it available queue* in tws but we get no / aren't processing wtv event comes back from that. imo we should just cancel orders like this and expect the user to read the error on console for now.

### DIFF
--- a/piker/brokers/ib.py
+++ b/piker/brokers/ib.py
@@ -25,7 +25,10 @@ from contextlib import asynccontextmanager
 from dataclasses import asdict
 from datetime import datetime
 from functools import partial
-from typing import List, Dict, Any, Tuple, Optional, AsyncIterator
+from typing import (
+    List, Dict, Any, Tuple, Optional,
+    AsyncIterator, Awaitable,
+)
 import asyncio
 from pprint import pformat
 import inspect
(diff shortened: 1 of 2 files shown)


goodboy:
Also tossed in the draft config file stuff from way back which we need for the new position / order config account selection.

goodboy:
This gets adhoc contracts search working and speeds up the existing stocks search by dropping an extra request call that it seems wasn't strictly necessary.

piker/brokers/ib.py
```diff
@@ -279,50 +295,70 @@ async def bars(
...
-                    # won't IPC serialize
-                    as_dict.pop('secIdList')
-
-                    details[unique_sym] = as_dict
-
-                    if len(details) == upto:
-                        return details
-
-            return details
+                    # sometimes there's a weird extra suffix returned
```
goodboy, line 355: so dumb.

Issue #197, opened 2021-06-11
Feed statuses


goodboy:
Since #164, we've had multi-feed streaming for charts but it would sure be handy if you would get a overall *status* of a particular provider connection possibly in the new status bar introduced in #189 as well as possibly through some kind of aggregate feed widget eventually. Further it'd be handy to have a way to restart a provider if it goes down, seems slow, or has an error through some kind of manual control.

Couple ideas:
- offer a feed status widget *in* the status bar (kind like our mode label) that you can mouse interact with and possibly restart / respawn feeds through a context menu?
- offer a way to possibly re-route where a feed is draw from for down the road when we have aggregate data daemons running shared for a group of traders

More ideas welcome from the lurks!

PR #205, opened 2021-07-11
Ordermode pps

goodboy:
Working draft of highly game-like position and order tracking for order mode.

Target feature list:
- [x] pp off screen indicators on order mode activation
- [ ] the same for (groups of) order(s) per symbol -> this i'd like feedback on first.
- [x] paper engine positions support -> actually came in #218
  - [x] first in just memory and from scratch each app session -> NB: currently it's only working for the first loaded symbol (paper only) due to the way the `paperboi` engine is loaded with it's feeds; we need a fix for this.
  - [ ] eventually stored in `positions.toml`? Probably want to coordinate with #214.
- [x] $positioning for default orders
- [x] % pnl on pp indicator label
- [x] initial order mode info menu with selectable position style/sizing options

### DIFF
--- a/piker/brokers/ib.py
+++ b/piker/brokers/ib.py
@@ -196,6 +196,8 @@ _adhoc_futes_set = {
     'mgc.nymex',
 
     'xagusd.cmdty',  # silver spot
+    'ni.nymex', # silver futes
+    'qi.comex', # mini-silver futes
 }
 
 # exchanges we don't support at the moment due to not knowing
--- a/piker/brokers/ib.py
+++ b/piker/brokers/ib.py
@@ -1295,10 +1295,14 @@ def pack_position(pos: Position) -> dict[str, Any]:
     else:
         symbol = con.symbol
 
+    symkey = '.'.join([
+        symbol.lower(),
+        (con.primaryExchange or con.exchange).lower(),
+    ])
     return BrokerdPosition(
         broker='ib',
         account=pos.account,
-        symbol=symbol,
+        symbol=symkey,
         currency=con.currency,
         size=float(pos.position),
         avg_price=float(pos.avgCost) / float(con.multiplier or 1.0),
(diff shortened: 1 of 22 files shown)


piker/brokers/ib.py
```diff
@@ -1295,10 +1297,14 @@ def pack_position(pos: Position) -> dict[str, Any]:
     else:
         symbol = con.symbol
 
+    symkey = '.'.join([
+        symbol.lower(),
+        (con.primaryExchange or con.exchange).lower(),
```
goodboy, line 1302: This fixes a bug where on pp loading from the broker, a sub-string of the actual intended instrument symbol would match showing a pp indicator for the wrong chart. explicitly truncating and appending the exchange solves this.

piker/data/_sampling.py
```diff
@@ -233,16 +238,23 @@ async def sample_and_broadcast(
...
 
                 try:
-                    if tick_throttle:
-                        # this is a send mem chan that likely
-                        # pushes to the ``uniform_rate_send()`` below.
-                        await stream.send(quote)
-
-                    else:
-                        await stream.send({sym: quote})
+                    with trio.move_on_after(0.2) as cs:
```
goodboy, line 245: This is effectively a backpressure detector and should basically never get entered unless there's an issue on the consumer side (which hopefully all these are resolved now).

Not sure if we should put this onto master though with the `.breakpoint()` but I haven't seen it happen yet in practice.

piker/ui/_chart.py
```diff
@@ -232,8 +261,50 @@ def load_symbol(
...
+        ala "view mode".
+
+        '''
+        # go back to view-mode focus (aka chart focus)
+        self.clearFocus()
+        self.linkedsplits.chart.setFocus()
+
 
-class LinkedSplits(QtWidgets.QWidget):
+class ChartnPane(QFrame):
```
goodboy, line 274: Composite to make the order pane stuff work alongside the linked split charts.

goodboy:
Follow up issue list inspired or drafted in this patch set:
- feed statuses / labels in #197
- `accounts.<broker>` support and docs in `brokers.toml` as #219
- further allocator refinements for pp limiting as per #220
- wrapping fsp streams as `Feed`s in #216 which should lead to improvements for #192

PR #208, opened 2021-07-21
Fix a few QtGui -> QtWidget issues

hengestone:
Possibly caused by `pyqtgraph`

### DIFF
--- a/piker/ui/_chart.py
+++ b/piker/ui/_chart.py
@@ -19,7 +19,7 @@ High level Qt chart widgets.
 
 """
 import time
-from typing import Tuple, Dict, Any, Optional, List
+from typing import Tuple, Dict, Any, Optional
 from types import ModuleType
 from functools import partial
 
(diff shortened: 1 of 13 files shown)


hengestone:
Will update this later

goodboy:
@hengestone give this a rebase onto `master` if you don't mind. #209 should let the CI run now?

piker/ui/_chart.py
```diff
@@ -19,11 +19,11 @@
 
 """
 import time
-from typing import Tuple, Dict, Any, Optional
+from typing import Tuple, Dict, Any, Optional, List
```
goodboy, line 22: in 3.9 I'm pretty sure we don't need this?

goodboy (commented):
Yeah other then not needing the `typing.List` import any more for python 3.9 (which I think we've now pinned to) this looks good to me 😎

goodboy:
Nice it worked.

piker/ui/_chart.py
hengestone: OK, removed in `afafe56` below.

goodboy (approved):
@hengestone just waiting on feedback from one user who needs this same patch to get up and running.

Once we get the ok on that I'll merge this 🏄🏼

PR #212, opened 2021-08-10
Feed caching

goodboy:
Initial data feed caching over `piker.data.feed.open_feed()` using the new `maybe_open_feed()`.

Adds a new `_cacheables.py` which contains a bunch of helpers for cache-y things.
This relies on https://github.com/goodboy/tractor/pull/229 in order to give multiple actor-local task consumers broadcast access to quote streams.

Putting this up to get eyes on it and see if there's any reason not to start building streaming apis under this paradigm.

Still TODO:
- [x] we should also cache the [`Feed.index_stream()`](https://github.com/pikers/piker/blob/master/piker/data/feed.py#L368) per actor, which makes me also wonder if we should provide new `Feed` instances on cache hits? ~or can we just wrap it and override `._index_stream` or is that to *mutate*-y~ ended up using the new `maybe_cache_ctx()` mngr introduced in this PR
- attaching to a data feed currently [registers the client stream](https://github.com/pikers/piker/blob/master/piker/data/feed.py#L325) with the [sample and broadcast loop](https://github.com/pikers/piker/blob/96b0102f04fc12eae16ada20983f7c37e3c320c3/piker/data/_sampling.py#L234); would there be a benefit of switching from this `brokerd`-always-pushes to a *tasks always pull* model here?
  - i'm thinking *no*
  - the only thing that might be handy is a re-impl of [`uniform_rate_send`](https://github.com/pikers/piker/blob/96b0102f04fc12eae16ada20983f7c37e3c320c3/piker/data/_sampling.py#L258) as a `broadcast_receiver()` potentially? in which case i think we can drop all the timing logic and just let sent quotes queue up and then pull on a fixed period? i may be thinking about this wrong..
- [x] feed "pausing" support which allows sending a `pause`/`resume` message to the endpoint to add remove the subscription dynamically
- [ ] ~taking the `Feed` api to our fsp subsystem (this will likely get delayed to a new PR)~ -> #216


piker/_cacheables.py
```diff
@@ -0,0 +1,203 @@
...
+
+
+_cache: dict[str, 'Client'] = {}  # noqa
+
+
+# XXX: this mis mostly an alt-implementation of
+# maybe_open_ctx() below except it uses an async exit statck.
+# ideally wer pick one or the other.
+@asynccontextmanager
+async def open_cached_client(
```
goodboy, line 84: this is old code that i think should be ported to the `maybe_open_ctx()` below but for broker backend client instance.

piker/_cacheables.py
```diff
@@ -0,0 +1,203 @@
...
+    '''Maybe open a context manager if there is not already a cached
+    version for the provided ``key``. Return the cached instance on
+    a cache hit.
+
+    '''
+    @contextmanager
+    def get_and_use() -> AsyncIterable[T]:
+        # key error must bubble here
+        feed = cache.ctxs[key]
+        log.info(f'Reusing cached feed for {key}')
```
goodboy, line 160: i guess this should be some representation format instead of "feed" 😂

piker/data/feed.py
```diff
@@ -530,3 +526,39 @@ async def open_feed(
...
+            brokername,
+            [sym],
+            loglevel=loglevel,
+        ),
+    ) as (cache_hit, feed):
+
+        if cache_hit:
+            # add a new broadcast subscription for the quote stream
+            # if this feed is likely already in use
+            async with feed.stream.subscribe() as bstream:
```
goodboy, line 561: This is the part that requires https://github.com/goodboy/tractor/pull/229's `tokio` style broadcasting.

goodboy:
Just updated [`infect_asyncio`](https://github.com/goodboy/tractor/tree/infect_asyncio) in `tractor` to match. so we shud be guuuudd.

goodboy:
Probably also keeping 👀 on down the road is the [cachetools](https://github.com/tkem/cachetools) project.

piker/_cacheables.py
```diff
@@ -0,0 +1,203 @@
...
+        with get_and_use() as feed:
+            yield True, feed
+    except KeyError:
+        # lock feed acquisition around task racing  / ``trio``'s
+        # scheduler protocol
+        await cache.lock.acquire()
+        try:
+            with get_and_use() as feed:
+                cache.lock.release()
+                yield feed
```
goodboy, line 180: bug, should be `tuple[bool, T]`

goodboy:
Added `Feed.pause()`/`.resume()` so we can use it when charts are switched to avoid `brokerd` pushing more streams then necessary and also unseen draw cycles on charts not in focus 😎

piker/data/feed.py
```diff
@@ -322,11 +325,28 @@ async def attach_feed_bus(
...
+        subs = bus._subscribers[symbol]
+        subs.append(sub)
 
         try:
-            await trio.sleep_forever()
-
+            uid = ctx.chan.uid
+            fqsn = f'{symbol}.{brokername}'
+
+            async for msg in stream:
```
goodboy, line 335: The beauty of 2 way streamzz 🏄🏼‍♀️

piker/_cacheables.py
```diff
@@ -0,0 +1,222 @@
...
+    a cache hit.
+
+    '''
+
+    await cache.lock.acquire()
+
+    ctx_key = id(mngr)
+
+    # TODO: does this need to be a tractor "root nursery"?
+    async with maybe_open_nursery(cache.nurseries.get(ctx_key)) as n:
```
goodboy, line 184: I'm still slightly unclear how the teardown part of lifetime works here; pretty sure it's going to tear down when the *first* consumer task is complete instead of when the *last* arriving consumer does.

In order to get the latter behavior we might need to have an *actor global* nursery that's brought up with the runtime / the consumer process?

Not sure this absolutely must be addressed right now since usually the creator task stays up as long as the app / daemon which is using the feed.

piker/data/feed.py
```diff
@@ -335,6 +355,31 @@ async def attach_feed_bus(
             bus._subscribers[symbol].remove(sub)
 
 
+@asynccontextmanager
+async def open_sample_step_stream(
```
goodboy, line 359: Allows us to actor-cache the OHLC step event stream per delay (since you likely will want the same event for all local consumers.

piker/data/feed.py
```diff
@@ -530,3 +562,39 @@ async def open_feed(
...
+    '''
+    sym = symbols[0].lower()
+
+    async with maybe_open_ctx(
+        key=(brokername, sym),
+        mngr=open_feed(
+            brokername,
+            [sym],
+            loglevel=loglevel,
+            **kwargs,
```
goodboy, line 590: This `kwargs* pass through was critical...

guilledk (approved):
Love to see the new `_broadcast` machinery. Also `_cacheables` looking sexy!

iamzoltan (approved):
Looks kosher to me. Digging the improvements!

piker/_cacheables.py
goodboy: boo yeah. worked like a charm 🏄🏼

piker/_cacheables.py
```diff
@@ -0,0 +1,193 @@
...
+            # if no more consumers, teardown the client
+            if cache.users <= 0:
+                log.warning(f'De-allocating feed for {key}')
+
+                # terminate mngr nursery
+                cache.no_more_users.set()
+
+
+@asynccontextmanager
+async def open_cached_client(
```
goodboy, line 180: not bad right?

piker/_cacheables.py
```diff
@@ -0,0 +1,193 @@
...
+
+
+class cache:
+    '''Globally (processs wide) cached, task access to a
+    kept-alive-while-in-use async resource.
+
+    '''
+    lock = trio.Lock()
+    users: int = 0
+    values: dict[tuple[str, str], tuple[AsyncExitStack, Any]] = {}
```
goodboy, line 89: woops, stack is tossed now.

Issue #201, opened 2021-06-18
Let's get kraken (🙄 ), on orders..


goodboy:
As mentioned, we need new and improved `kraken` support for orders after the changes to the clearing system to include strict typed messaging semantics as per #190. There was ad-hoc support for this before that PR landed but the code is imo not usuable and was always a bit hacky by fitting too much to the `ib` backend schema that was in place (due to it being the first broker with full live trading support).

This issue is more or less a duplicate of #194 but for [our `kraken` backend](https://github.com/pikers/piker/blob/master/piker/brokers/kraken.py) so please use the detailed resources in that issue as a reference as well (and vice versa if anything useful get's stuck here @guilledk).

---
The technical details of how to implement the 2 way streaming endpoint `trades_dialogue()` is in the first bullet in #194 and both the paper engine and current `ib` backend implementation should be used as reference.

- before you do anything else you're going to have to add [authentication code](https://docs.kraken.com/rest/#section/Authentication) to connect to a user's account and get an api key.
  - private user api tokens/data for this will ideally be stored in our [`brokers.toml`](https://github.com/pikers/piker/blob/master/data/brokers.toml) under a `kraken` section (i no joke, somehow lost the schema i had originally in this file bc i was lazy and forgot to put it in git..)
  - make sure we're adhering to the api's [rate limit](https://docs.kraken.com/rest/#section/Rate-Limits) constraints
- the first bit to relay from the `trades_dialogue()` endpoint is the open set of positions likely from the [account balance](https://docs.kraken.com/rest/#operation/getAccountBalance) ReST endpoint (so **not** websockets)
  - this data can be related back to the EMS as a sequence of [`BrokerdPosition` messages](https://github.com/pikers/piker/blob/master/piker/clearing/_messages.py#L227) as is [done in the `ib` case](https://github.com/pikers/piker/blob/master/piker/brokers/ib.py#L1302)
- **NOTE**: there's 2 options for sending order requests and not sure which one is going to be better (i had previously done websockets but the rest approach might be a lot simpler and more deterministic?):
  - using the rest api: https://docs.kraken.com/rest/#tag/User-Trading
  - using the websocket api: https://docs.kraken.com/websockets/
    - be sure to read the *general considerations* there.
- next, you'll likely want to dig into the [websocket api](https://docs.kraken.com/websockets) to leverage the real-time streaming for sending updates to our EMS:
  - live trades endpoint: https://docs.kraken.com/websockets/#message-ownTrades
[...]

PR #213, opened 2021-08-24
Brokers config schema bump, ib patches

goodboy:
To aid with #201 for the auth step and storing keys locally in the filesystem.

This includes some patches for the `ib` backend that have been sitting in dev branches.

Nothing too cray here.


iamzoltan (approved):
looks good. Thanks for the broker.toml template!

PR #214, opened 2021-08-26
Phil MacKraken 

iamzoltan:
Working on adding order support and position calculation for the kraken backend. It's definitely a work in progress. Any comments would be appreciated. @goodboy @guilledk

### DIFF
--- a/piker/brokers/kraken.py
+++ b/piker/brokers/kraken.py
@@ -40,7 +40,10 @@ from ._util import resproc, SymbolNotFound, BrokerError
 from ..log import get_logger, get_console_log
 from ..data import ShmArray
 from ..data._web_bs import open_autorecon_ws
-from ..clearing._messages import BrokerdPosition, BrokerdOrder, BrokerdStatus
+from ..clearing._messages import (
+    BrokerdPosition, BrokerdOrder, BrokerdStatus,
+    BrokerdOrderAck, BrokerdError, BrokerdCancel, BrokerdFill
+)
 
 import urllib.parse
 import hashlib
@@ -433,6 +436,74 @@ def normalize_symbol(
         return ticker.lower()
 
 
+async def handle_order_requests(
+
+        client: #kraken,
+        ems_order_stream: tractor.MsgStream,
+
+) -> None:
+
+    # order_request: dict
+    async for request_msg in ems_order_stream:
+        log.info(f'Received order request {request_msg}')
+
+        action = request_msg['action']
+
+        if action in {'buy', 'sell'}:
+
+            account = request_msg['account']
+            if account != 'kraken.spot':
+                log.error(
+                    'This is a kraken account, \
+                    only a `kraken.spot` selection is valid'
+                )
+                await ems_order_stream.send(BrokerError(
+                    oid=request_msg['oid']
+                    symbol=request_msg['symbol']
+                    reason=f'Kraken only, No account found: `{account}` ?',
+                ).dict())
+                continue
+
+            # validate
+            order = BrokerdOrder(**request_msg)
+
+            # call our client api to submit the order
+            ## TODO: look into the submit_limit method, do it write my own?
+            reqid = await client.submit_limit(
+
+                oid=order.oid,
+                symbol=order.symbol,
+                price=order.price,
+                action=order.action,
+                size=order.size,
+                ## XXX: how do I handle new orders
+                reqid=order.reqid,
+            )
+
+            # deliver ack that order has been submitted to broker routing
+            await ems_order_stream.send(
+                BrokerdOrderAck(
+
+                    # ems order request id
+                    oid=order.oid,
+
+                    # broker specific request id
+                    reqid=reqid,
+
+                ).dict()
+            )
+
+        elif action == 'cancel':
+             msg = BrokerdCancel(**request_msg)
+
+             await client.submit_cancel(
+                reqid=msg.reqid
+            )
+
+        else:
+            log.error(f'Unknown order command: {request_msg}')
+
+
 @tractor.context
 async def trades_dialogue(
     ctx: tractor.Context,
@@ -462,9 +533,10 @@ async def trades_dialogue(
 
         await trio.sleep_forever()
 
-   #     async with (
-   #         ctx.open_stream() as ems_stream,
-   #         
+       # async with (
+       #     ctx.open_stream() as ems_stream,
+       #     trio.open_nursery as n,
+       # ):
 
 
 async def stream_messages(ws):
--- a/piker/brokers/kraken.py
+++ b/piker/brokers/kraken.py
@@ -20,7 +20,7 @@ Kraken backend.
 """
 from contextlib import asynccontextmanager
 from dataclasses import asdict, field
-from typing import List, Dict, Any, Tuple, Optional
+from typing import List, Dict, Any, Tuple, Optional, AsyncIterator
 import time
 
 from trio_typing import TaskStatus
@@ -34,12 +34,13 @@ from pydantic.dataclasses import dataclass
 from pydantic import BaseModel
 import wsproto
 
-from . import config
+from .. import config
 from .._cacheables import open_cached_client
 from ._util import resproc, SymbolNotFound, BrokerError
 from ..log import get_logger, get_console_log
 from ..data import ShmArray
 from ..data._web_bs import open_autorecon_ws
+from ..clearing._messages import BrokerdPosition
 
 import urllib.parse
 import hashlib
@@ -229,41 +230,35 @@ class Client:
             print(err)
         return resp['result']
 
-    async def get_ledger(
+    async def get_positions(
         self,
-        data: Dict[str, Any]
-        ) -> pd.DataFrame:
-        ledgers = await self.get_user_data('Ledgers', data)
-        num_entries = int(ledgers['count'])
-        crypto_transactions = np.empty((num_entries, 5), dtype=object)
-        if num_entries // 50 < 0 or num_entries == 50:
-        # NOTE: Omitting the following values from the kraken ledger:
-        #            -> 'refid', 'type', 'subtype', 'aclass'
-            for i, entry in enumerate(ledgers['ledger'].items()):
-                crypto_transactions[i] = [
-                    entry[1]['time'],
-                    entry[1]['amount'],
-                    entry[1]['fee'],
-                    entry[1]['asset'],
-                    entry[1]['balance']
-                ]
-        else:
-            for n in range(num_entries // 50 + 1):
-                data['ofs'] = n * 50
-                ledgers = await self.get_user_data('Ledgers', data)
-                for i, entry in enumerate(ledgers['ledger'].items()):
-                        crypto_transactions[i + n * 50] = [
-                            entry[1]['time'],
-                            entry[1]['amount'],
-                            entry[1]['fee'],
-                            entry[1]['asset'],
-                            entry[1]['balance']
-                        ]
-        ledger = pd.DataFrame(
-            columns = ['time', 'amount', 'fee', 'asset', 'balance'],
-            data = crypto_transactions
-        )
-        return ledger
+        data: Dict[str, Any] = {}
+    ) -> Dict[str, Any]:
+        balances = await self.get_user_data('Balance', data)
+        ## TODO: grab all entries, not just first 50
+        traders = await self.get_user_data('TradesHistory', data)
+        positions = {}
+        vols = {}
+        
+        # positions
+        ## TODO: Make sure to add option to include fees in positions calc
+        for trade in traders['trades'].values():
+            sign = -1 if trade['type'] == 'sell' else 1
+            try:
+                positions[trade['pair']] += sign * float(trade['cost'])
+                vols[trade['pair']] += sign * float(trade['vol'])
+            except KeyError:
+                positions[trade['pair']] = sign * float(trade['cost'])
+                vols[trade['pair']] = sign * float(trade['vol'])
+    
+        for pair in positions.keys():
+            asset_balance = vols[pair]
+            if asset_balance == 0:
+                positions[pair] = 0
+            else:
+                positions[pair] /= asset_balance
+
+        return positions
 
     async def symbol_info(
         self,
@@ -385,80 +380,9 @@ async def get_client() -> Client:
     data = {
         # add non-nonce and non-ofs vars
     }
+    # positions = await client.get_positions(data)
 
-    balances = await client.get_user_data('Balance', data)
-    traders = await client.get_user_data('TradesHistory', data)
-    ledger = await client.get_ledger(data)
-    
-    # positions
-    ## TODO: Make sure to add option with fees
-    n, m = ledger.shape
-    ledger['time'] = ledger['time'].apply(lambda x: int(x))
-    assets = set(ledger['asset'])
-    trade_times = set(ledger['time'])
-    trades = {}
-    positions = {}
-    # for index, row in ledger.iterrows():
-    #     if index == n:
-    #         break
-    #     asset = row['asset']
-    #     ## TODO: Look into a way to generalize this
-    #     if asset != 'ZEUR' and ledger.loc[index+1, 'asset'] == 'ZEUR':
-    #         try:
-    #             trades[asset]['amounts'].append(float(ledger.loc[index, 'amount']))
-    #             trades[asset]['prices'].append(float(ledger.loc[index+1, 'amount']))
-    #         except KeyError:
-    #             trades[asset] = {
-    #                 'amounts': [float(ledger.loc[index+1, 'amount'])],
-    #                 'prices': [float(ledger.loc[index+1, 'amount'])]
-    #             }
-
-    ## TODO: Look into a way to generalize this for any fiat
-    ## TODO: Figure out how to handle coin for coin trades
-    for trade_time in trade_times:
-        trade = ledger[ledger['time'] == trade_time]
-        coin = trade[trade['asset'] != 'ZEUR']
-        fiat = trade[trade['asset'] == 'ZEUR']
-        if len(coin) == 0 or len(coin) > 1:
-            continue
-        asset = list(coin.loc[:, 'asset'])[0]
-        amount = np.sum(coin['amount'].apply(lambda x: float(x)))
-        if amount > 0:
-            sign = -1
-            price = sign * np.sum(fiat['amount'].apply(lambda x: float(x)))
-            try:
-                trades[asset]['trade_amounts'].append(amount)
-                trades[asset]['trade_prices'].append(price)
-                trades[asset]['enter_amounts'].append(amount)
-            except KeyError:
-                trades[asset] = {
-                    'trade_amounts': [amount],
-                    'trade_prices': [price],
-                    'enter_amounts': [amount]
-                }
-        else:
-            price = 0
-            # continue
-            try:
-                trades[asset]['trade_amounts'].append(amount)
-                trades[asset]['trade_prices'].append(price)
-            except KeyError:
-                trades[asset] = {
-                    'trade_amounts': [amount],
-                    'trade_prices': [price],
-                    'enter_amounts': []
-                }
-
-
-    for asset in assets:
-        if asset == 'ZEUR':
-            continue
-        t_amounts = np.array(trades[asset]['trade_amounts'])
-        t_prices = np.array(trades[asset]['trade_prices'])
-        e_amounts = np.array(trades[asset]['enter_amounts'])
-        positions[asset] = np.dot(np.divide(t_prices, t_amounts), t_amounts) / np.sum(e_amounts)
-
-    await tractor.breakpoint()
+    # await tractor.breakpoint()
 
     # at startup, load all symbols locally for fast search
     await client.cache_symbols()
@@ -466,27 +390,32 @@ async def get_client() -> Client:
     yield client
 
 
-# @tractor.context
-# async def trades_dialogue(
-#     ctx: tractor.Context,
-#     loglevel: str = None,
-# ) -> AsyncIterator[Dict[str, Any]]:
-# 
-#     # XXX: required to propagate ``tractor`` loglevel to piker logging
-#     get_console_log(loglevel or tractor.current_actor().loglevel)
-# 
-#     # deliver positions to subscriber before anything else
-#     positions = await _trio_run_client_method(method='positions')
-# 
-#     all_positions = {}
-# 
-#     for pos in positions:
-#         msg = pack_position(pos)
-#         all_positions[msg.symbol] = msg.dict()
-# 
-#     await ctx.started(all_positions)
+@tractor.context
+async def trades_dialogue(
+    ctx: tractor.Context,
+    loglevel: str = None,
+) -> AsyncIterator[Dict[str, Any]]:
+
+    # XXX: required to propagate ``tractor`` loglevel to piker logging
+    get_console_log(loglevel or tractor.current_actor().loglevel)
+
+    # deliver positions to subscriber before anything else
+    # positions = await _trio_run_client_method(method='positions')
+
+    global _accounts2clients
+
+    positions = await client.get_positions()
+
+    await tractor.breakpoint()
 
 
+    all_positions = {}
+
+    for pos in positions:
+        msg = pack_position(pos)
+        all_positions[msg.symbol] = msg.dict()
+
+    await ctx.started(all_positions)
 
 
 async def stream_messages(ws):
--- a/piker/brokers/kraken.py
+++ b/piker/brokers/kraken.py
@@ -243,18 +243,17 @@ class Client:
         uri_path = f'/0/private/{method}'
         data['nonce'] = str(int(1000*time.time()))
         resp = await self._private(method, data, uri_path)
-        err = resp['error']
-        if err:
-            print(err)
-        return resp['result']
+        return resp
 
     async def get_positions(
         self,
         data: Dict[str, Any] = {}
     ) -> Dict[str, Any]:
-        balances = await self.kraken_endpoint('Balance', data)
+        resp = await self.kraken_endpoint('Balance', data)
+        balances = resp['result']
         ## TODO: grab all entries, not just first 50
-        traders = await self.kraken_endpoint('TradesHistory', data)
+        resp = await self.kraken_endpoint('TradesHistory', data)
+        traders = resp['result']
         positions = {}
         vols = {}
         
@@ -284,7 +283,7 @@ class Client:
         symbol: str,
         price: float,
         action: str,
-        size: str,
+        size: float,
 #        account: str,
         reqid: int = None,
     ) -> int:
@@ -296,14 +295,26 @@ class Client:
             "userref": reqid,
             "ordertype": "limit",
             "type": action,
-            "volume": size,
+            "volume": str(size),
             "pair": symbol,
-            "price": price,
-            "validate": True
+            "price": str(price),
+            # set to True test AddOrder call without a real submission
+            "validate": False
         }
         resp = await self.kraken_endpoint('AddOrder', data)
+        return resp
+
+    async def submit_cancel(
+        self,
+        reqid: str,
+    ) -> None:
+        """Send cancel request for order id ``reqid``.
+
+        """
+        # txid is a transaction id given by kraken
+        data = {"txid": reqid}
+        resp = await self.kraken_endpoint('CancelOrder', data)
         print(resp)
-        return reqid
 
     async def symbol_info(
         self,
@@ -495,10 +506,11 @@ async def handle_order_requests(
                 continue
 
             # validate
+            temp_id = next(userref_counter)
             order = BrokerdOrder(**request_msg)
 
             # call our client api to submit the order
-            reqid = await client.submit_limit(
+            resp = await client.submit_limit(
 
                 oid=order.oid,
                 symbol=order.symbol,
@@ -506,21 +518,40 @@ async def handle_order_requests(
                 action=order.action,
                 size=order.size,
                 ## XXX: how do I handle new orders
-                reqid=next(userref_counter),
+                reqid=temp_id,
             )
 
-            # deliver ack that order has been submitted to broker routing
-            await ems_order_stream.send(
-                BrokerdOrderAck(
+            err = resp['error']
+            if err:
+                log.error(f'Failed to submit order')
+                await ems_order_stream.send(
+                    BrokerdError(
+                        oid=order.oid,
+                        reqid=temp_id,
+                        symbol=order.symbol,
+                        reason="Failed order submission",
+                        broker_details=resp
+                    ).dict()
+                )
+            else:
+                ## TODO: handle multiple cancels
+                ##       txid is an array of strings
+                reqid = resp['result']['txid'][0]
+                # deliver ack that order has been submitted to broker routing
+                await ems_order_stream.send(
+                    BrokerdOrderAck(
 
-                    # ems order request id
-                    oid=order.oid,
+                        # ems order request id
+                        oid=order.oid,
 
-                    # broker specific request id
-                    reqid=reqid,
+                        # broker specific request id
+                        reqid=reqid,
 
-                ).dict()
-            )
+                        # account the made the order
+                        account=order.account
+
+                    ).dict()
+                )
 
         elif action == 'cancel':
              msg = BrokerdCancel(**request_msg)
--- a/piker/brokers/kraken.py
+++ b/piker/brokers/kraken.py
@@ -33,6 +33,7 @@ import tractor
 from pydantic.dataclasses import dataclass
 from pydantic import BaseModel
 import wsproto
+from itertools import count
 
 from .. import config
 from .._cacheables import open_cached_client
@@ -290,9 +291,9 @@ class Client:
         """Place an order and return integer request id provided by client.
 
         """
-        # Build order data from kraken
+        # Build order data for kraken api
         data = {
-            "userref": 1,
+            "userref": reqid,
             "ordertype": "limit",
             "type": action,
             "volume": size,
@@ -465,12 +466,14 @@ def normalize_symbol(
 
 async def handle_order_requests(
 
-        client: 'test',#kraken,
+        client: Client,
         ems_order_stream: tractor.MsgStream,
 
 ) -> None:
 
-    # order_request: dict
+    request_msg: dict
+    order: BrokerdOrder
+    userref_counter = count()
     async for request_msg in ems_order_stream:
         log.info(f'Received order request {request_msg}')
 
@@ -484,7 +487,7 @@ async def handle_order_requests(
                     'This is a kraken account, \
                     only a `kraken.spot` selection is valid'
                 )
-                await ems_order_stream.send(BrokerError(
+                await ems_order_stream.send(BrokerdError(
                     oid=request_msg['oid'],
                     symbol=request_msg['symbol'],
                     reason=f'Kraken only, No account found: `{account}` ?',
@@ -503,7 +506,7 @@ async def handle_order_requests(
                 action=order.action,
                 size=order.size,
                 ## XXX: how do I handle new orders
-                reqid=order.reqid,
+                reqid=next(userref_counter),
             )
 
             # deliver ack that order has been submitted to broker routing
--- a/piker/brokers/kraken.py
+++ b/piker/brokers/kraken.py
@@ -34,11 +34,18 @@ from pydantic.dataclasses import dataclass
 from pydantic import BaseModel
 import wsproto
 
+from . import config
 from .._cacheables import open_cached_client
 from ._util import resproc, SymbolNotFound, BrokerError
 from ..log import get_logger, get_console_log
 from ..data import ShmArray
 from ..data._web_bs import open_autorecon_ws
+
+import urllib.parse
+import hashlib
+import hmac
+import base64
+
 
 log = get_logger(__name__)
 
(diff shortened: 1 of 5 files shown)


piker/brokers/kraken.py
```diff
@@ -119,6 +119,19 @@ class Pair(BaseModel):
     ordermin: float  # minimum order volume for pair
 
 
+class Order(BaseModel):
```
goodboy, line 122: I would probably consider trying to type-validate these where possible though it's not a huge deal yet.

piker/brokers/kraken.py
```diff
@@ -523,50 +538,65 @@ async def handle_order_requests(
...
-
-                        # broker specific request id
-                        reqid=reqid,
-
-                        # account the made the order
-                        account=order.account
-
-                    ).dict()
-                )
+            #resp = await client.submit_limit(
```
goodboy, line 559: Yeah, i don't this you have to necessarily switch to the `ws` version here if the http api worked well already (which it did when i tested it), but hey great if we want both and or the ws one does work better.

brining in the ws support could also come in a follow up issue + PR 😺

piker/brokers/kraken.py
```diff
@@ -523,50 +538,65 @@ async def handle_order_requests(
...
+                          "ordertype": "limit",
+                          "pair": slashinsert(order.symbol.upper()),
+                          "price": str(order.price),
+                          "token": token,
+                          "type": order.action,
+                          "volume": str(order.size)
+                        }
+                    
+
+            await ws.send_msg(order_msg)
```
goodboy, line 556: Ok so i guess from kraken docs you're supposed to get pub messages back on the `ws` of the form:

```json
{
  "descr": "buy 0.01770000 XBTUSD @ limit 4000",
  "event": "addOrderStatus",
  "status": "ok",
  "txid": "ONPNXH-KMKMU-F4MR5V"
}
```
i'm seeing from [here](https://docs.kraken.com/websockets/#message-addOrder).

I'll try testing this in a bit.

piker/brokers/kraken.py
```diff
@@ -801,6 +846,29 @@ async def process_order_msgs(
         # HINT: create a ``pydantic.BaseModel`` to parse and validate
         # and then in the caller recast to our native ``BrokerdX`` msg types.
 
+        # check that we are on openOrders stream XXX: Maybe do something with sequence
```
goodboy, line 849: i wouldn't put this here. this is exactly why i added the `process_order_msgs()` stream handler in the previous commit.

piker/brokers/kraken.py
```diff
@@ -801,6 +846,29 @@ async def process_order_msgs(
...
+        assert msg[1] == 'openOrders'
+        orders = msg[0]
+
+        for order in orders:
+            txid = list(order.keys())[0]
+            order_msg = Order(
```
goodboy, line 855: yah looks gud 👍🏼

piker/brokers/kraken.py
goodboy: yeah these should also show up in the `process_order_msgs()` handler below *i think?*

piker/brokers/kraken.py
```diff
@@ -685,16 +715,31 @@ async def subscribe(ws: wsproto.WSConnection, token: str):
             trio.open_nursery() as n,
         ):
             ## TODO: maybe add multiple accounts
-            n.start_soon(handle_order_requests, client, ems_stream)
+            #n.start_soon(handle_order_requests, client, ems_stream)
```
goodboy, line 718: making sure you've noticed this.

as it stands if you don't relay the order request of course you can't expect to get status updates 😂

piker/brokers/kraken.py
```diff
@@ -716,16 +724,70 @@ async def subscribe(ws: wsproto.WSConnection, token: str):
...
+                    ws,
+                    token,
+                    userref_oid_map
+                )
                 from pprint import pprint
+                pending_orders = []
                 async for msg in process_order_msgs(ws):
                     pprint(msg)
+                    for order in msg:
+                        ## TODO: Maybe do a better
```
goodboy: do a better what? 😉

piker/data/_web_bs.py
```diff
@@ -81,9 +83,15 @@ async def _connect(
                     trio_websocket.open_websocket_url(self.url)
                 )
                 # rerun user code fixture
-                ret = await self._stack.enter_async_context(
-                    self.fixture(self)
-                )
+                if self.token == '':
```
goodboy, line 86: why not just always pass the token in?
i would presume all web backends will need it.

piker/brokers/kraken.py
goodboy: It would be pretty cool to test the latency different between these.
I got a feeling we might also not run into as many rate limiting problems on the `kraken` side with ws, but it's probably worth testing.

piker/brokers/kraken.py
iamzoltan: I think I do this somewhere

goodboy: `pydantic` will do it faster in C extensions too.

piker/brokers/kraken.py
iamzoltan: performance aside, using the http api solves the issue of having to keep track of mappings, which I think becomes tricky if you have existing orders that reference the same `userref`.

That being said, the `ws` exercise was good and the code can easily be repurposed for the `ownTrades` stream to keep track of successful trades.

Thoughts? @goodboy @guilledk

piker/brokers/kraken.py
iamzoltan: these dont come in if you are subscribed. Thats the response you would get if you just made the request. You get a version of that information in the format they line out in the description of the stream. The docs are really poorly written in that sense.

piker/brokers/kraken.py
iamzoltan: why do I feel like this is old code?

goodboy: because it is 😂 ?

look at the timestamps of the comments lol

piker/brokers/kraken.py
```diff
@@ -275,14 +438,321 @@ async def bars(
...
+
+                            action=trade.action,
+                            size=float(trade.size),
+                            price=float(trade.price),
+                            ## TODO: maybe capture more msg data i.e fees?
+                            broker_details={'name': 'kraken'},
+                            broker_time=float(trade.broker_time)
+                        )
+                        
+                        await ems_stream.send(fill_msg.dict())
```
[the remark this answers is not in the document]
goodboy: > so I had to add this to get the triangles to show. I tried its on its own, nothing (line remained, but no triangle), I tried just the status update (line gone, but no triangle), and then after digging into the code, I decided to try both, and that worked.

Ahh yeah could be. If that's the case then all good.

> fyi, I noticed in your comments you mentioned potentially getting rid of the BrokerdFill msg as it seems redundant from ib, this would break the triangles function of the ui as it needs both msgs to work.

Yeah then let's keep it for now. I think the `ib` case is weird as well because they send multiple messages for all sorts of stuff. For example there's like a separate msg for the commissions report. I think it also makes sense to have fill msgs be distinct since you can definitely have multiple fills *complete* and order execution depending on book liquidity.

guilledk (commented):
Looking great so far

piker/brokers/kraken.py
```diff
@@ -20,7 +20,7 @@
 """
 from contextlib import asynccontextmanager
 from dataclasses import asdict, field
-from typing import List, Dict, Any, Tuple, Optional
+from typing import List, Dict, Any, Tuple, Optional, AsyncIterator
```
goodboy, line 23: Drop all these data structures, we're on 3.9+ now so use the new built-in syntax.

piker/brokers/kraken.py
```diff
@@ -33,12 +33,25 @@
...
 from ..data import ShmArray
-from ..data._web_bs import open_autorecon_ws
+from ..data._web_bs import open_autorecon_ws, NoBsWs
+from ..clearing._messages import (
+    BrokerdPosition, BrokerdOrder, BrokerdStatus,
+    BrokerdOrderAck, BrokerdError, BrokerdCancel,
+    BrokerdFill,
+)
+
+import urllib.parse
```
goodboy, line 50: our current import style is to have:
- built-in modules (aka stdlib)
- 3rd party
- internal

so move these above the internals.

piker/brokers/kraken.py
```diff
@@ -106,6 +119,15 @@ class Pair(BaseModel):
     ordermin: float  # minimum order volume for pair
 
 
+class Trade(BaseModel):
+    """Trade class that helps parse and validate ownTrades stream"""
```
goodboy, line 123: I've started adopting a doc string format with `'''`.

I'm not sure if it'll still but for now I think i'd prefer to give it a try.
It includes a final blank line before the closing `'''`.

piker/brokers/kraken.py
```diff
@@ -106,6 +119,15 @@ class Pair(BaseModel):
...
+    reqid: str # kraken order transaction id
```
goodboy, line 124: If you're going to use `pydantic` you should try to type these to make it more purposeful. They even support a time format now iirc though not sure it'll work in this case.

piker/brokers/kraken.py
```diff
@@ -129,6 +151,41 @@ class OHLC:
...
 
 
+def get_config() -> dict[str, Any]:
+
+    conf, path = config.load()
+
+    section = conf.get('kraken')
+
+    if section is None:
+        log.warning(f'No config section found for kraken in {path}')
```
goodboy, line 161: so what happens if there is no section?
have you tested this case?

piker/brokers/kraken.py
```diff
@@ -129,6 +151,41 @@ class OHLC:
...
+        return {}
+
+    return section
+
+
+def get_kraken_signature(
+    urlpath: str,
+    data: Dict[str, Any],
```
goodboy, line 169: example of where you can just use `dict[str, Any]` now.

piker/brokers/kraken.py
```diff
@@ -129,6 +151,41 @@ class OHLC:
...
+    encoded = (str(data['nonce']) + postdata).encode()
+    message = urlpath.encode() + hashlib.sha256(encoded).digest()
+
+    mac = hmac.new(base64.b64decode(secret), message, hashlib.sha512)
+    sigdigest = base64.b64encode(mac.digest())
+    return sigdigest.decode()
+
+
+class InvalidKey(ValueError):
+    """EAPI:Invalid key
```
goodboy, line 182: Same thing with this doc str (i won't mention the rest) and also the paragraph here is indented weird no?

piker/brokers/kraken.py
```diff
@@ -162,6 +222,129 @@ async def _public(
...
+        return resproc(resp, log)
+
+    async def kraken_endpoint(
+        self,
+        method: str,
+        data: Dict[str, Any]
+    ) -> Dict[str, Any]:
+        uri_path = f'/0/private/{method}'
+        data['nonce'] = str(int(1000*time.time()))
+        resp = await self._private(method, data, uri_path)
```
goodboy, line 254: drop intermediary `resp`.

piker/brokers/kraken.py
```diff
@@ -162,6 +222,129 @@ async def _public(
...
+        }
+        resp = await self._sesh.post(
+            path=f'/private/{method}',
+            data=data,
+            headers=headers,
+            timeout=float('inf')
+        )
+        return resproc(resp, log)
+
+    async def kraken_endpoint(
```
goodboy, line 247: hmm, do we really need `kraken_` in this method name?
I think it's implied no?

piker/brokers/kraken.py
```diff
@@ -162,6 +222,129 @@ async def _public(
...
+    async def get_positions(
+        self,
+        data: Dict[str, Any] = {}
+    ) -> Dict[str, Any]:
+        data['ofs'] = 0
+        positions = {}
+        vols = {}
+        # Grab all trade history
+        while True:
+            resp = await self.kraken_endpoint('TradesHistory', data)
```
goodboy, line 266: add a comment that links to the API docs for this ep.

piker/brokers/kraken.py
```diff
@@ -162,6 +222,129 @@ async def _public(
...
+
+    async def get_positions(
+        self,
+        data: Dict[str, Any] = {}
+    ) -> Dict[str, Any]:
+        data['ofs'] = 0
+        positions = {}
+        vols = {}
+        # Grab all trade history
+        while True:
```
goodboy, line 265: not sure if there's some kind of state you're keeping here (`ofs` ?) but in theory you can do these requests concurrently and speedup the startup by doing a batch request for all trades (starting at different counts?) and then a single logic block to the aggregation calcs/work.

piker/brokers/kraken.py
```diff
@@ -162,6 +222,129 @@ async def _public(
...
+            elif resp['result']['trades'] == {}:
+                count = resp['result']['count']
+                break
+            # update existing dict if num trades exceeds 100
+            else: 
+                trades.update(resp['result']['trades'])
+            # increment the offset counter
+            data['ofs'] += 50
+            # To avoid exceeding API rate limit in case of a lot of trades
+            time.sleep(1)
```
goodboy, line 284: why?
never do this.

if anything you should be using `await trio.sleep(1)`

piker/brokers/kraken.py
goodboy: oh yeah looks like you keep passing `data` so i would figure out if you can do this batch-wise though it might be tricky to figure out *the end* trade given that you seem to be checking `resp['result']['trades'] == {}`?

piker/brokers/kraken.py
```diff
@@ -162,6 +222,129 @@ async def _public(
...
+        assert count == len(trades.values())
+
+        # positions
+        ## TODO: Make sure to add option to include fees in positions calc
+        for trade in trades.values():
+            sign = -1 if trade['type'] == 'sell' else 1
+            try:
+                positions[trade['pair']] += sign * float(trade['cost'])
+                vols[trade['pair']] += sign * float(trade['vol'])
+            except KeyError:
```
goodboy, line 296: i don't understand this. why catch a key error, you should explain this case with a detailed comment, and personally i'd avoid a catch if possible and use `dict.get()` instead + a logic block.

piker/brokers/kraken.py
```diff
@@ -162,6 +222,129 @@ async def _public(
...
+                positions[trade['pair']] = sign * float(trade['cost'])
+                vols[trade['pair']] = sign * float(trade['vol'])
+    
+        for pair in positions.keys():
+            asset_balance = vols[pair]
+            if asset_balance == 0:
+                positions[pair] = 0
+            else:
+                positions[pair] /= asset_balance
```
goodboy, line 305: describe in comment what this is computing.

piker/brokers/kraken.py
```diff
@@ -162,6 +222,129 @@ async def _public(
...
+
+        return positions, vols
```
goodboy, line 307: your return type annotation doesn't match this.

piker/brokers/kraken.py
```diff
@@ -162,6 +222,129 @@ async def _public(
...
+            "userref": reqid,
+            "ordertype": "limit",
+            "type": action,
+            "volume": str(size),
+            "pair": symbol,
+            "price": str(price),
+            # set to True test AddOrder call without a real submission
+            "validate": False
+        }
+        resp = await self.kraken_endpoint('AddOrder', data)
```
goodboy, line 333: again, drop the `resp`.

piker/brokers/kraken.py
```diff
@@ -162,6 +222,129 @@ async def _public(
...
+    async def submit_cancel(
+        self,
+        reqid: str,
+    ) -> None:
+        """Send cancel request for order id ``reqid``.
+
+        """
+        # txid is a transaction id given by kraken
+        data = {"txid": reqid}
+        resp = await self.kraken_endpoint('CancelOrder', data)
```
goodboy, line 345: again.

piker/brokers/kraken.py
```diff
@@ -162,6 +222,129 @@ async def _public(
...
+
+    async def submit_cancel(
+        self,
+        reqid: str,
+    ) -> None:
+        """Send cancel request for order id ``reqid``.
+
+        """
+        # txid is a transaction id given by kraken
+        data = {"txid": reqid}
```
goodboy, line 344: you can probably just pass this inline instead of the static `dict` declaration.

piker/brokers/kraken.py
```diff
@@ -271,18 +454,304 @@ async def bars(
             raise SymbolNotFound(json['error'][0] + f': {symbol}')
 
 
+
 @asynccontextmanager
 async def get_client() -> Client:
     client = Client()
 
+    section = get_config()
+    client._name = section['key_descr']
```
goodboy, line 463: This is, interesting. Why aren't these assigned in `.__init__()` which is more traditional style?

piker/brokers/kraken.py
```diff
@@ -275,14 +443,396 @@ async def bars(
...
+    #}
+
     # at startup, load all symbols locally for fast search
     await client.cache_symbols()
 
     yield client
 
 
-async def stream_messages(ws):
+def pack_position(
```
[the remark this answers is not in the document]
goodboy: Not totally sure i follow but it seem mostly benign for now.

piker/brokers/kraken.py
```diff
@@ -271,18 +454,304 @@ async def bars(
...
+        size=float(vol),
+        avg_price=float(pos),
+    )
+
+
+def normalize_symbol(
+    ticker: str
+) -> str:
+    symlen = len(ticker)
+    if symlen == 6:
```
goodboy, line 494: comment for this check, i presume it's the traditional 3x3 pairs symbol syntax?

piker/brokers/kraken.py
```diff
@@ -271,18 +454,304 @@ async def bars(
...
+        'event': 'subscribe',
+        'subscription': data,
+    }
+
+
+async def handle_order_requests(
+
+        client: Client,
+        ems_order_stream: tractor.MsgStream,
+        #ws: NoBsWs,
```
goodboy, line 522: drop these old commented values?

piker/brokers/kraken.py
```diff
@@ -271,18 +454,304 @@ async def bars(
...
+                    BrokerdError(
+                        oid=order.oid,
+                        reqid=temp_id,
+                        symbol=order.symbol,
+                        reason="Failed order submission",
+                        broker_details=resp
+                    ).dict()
+                )
+            else:
+                ## TODO: handle multiple cancels
```
goodboy, line 578: use pep8 style comment `#`

piker/brokers/kraken.py
```diff
@@ -271,18 +454,304 @@ async def bars(
...
+                            reqid=msg.reqid,
+                            account=msg.account,
+                            time_ns=time.time_ns(),
+                            status='cancelled',
+                            reason='Order cancelled',
+                            broker_details={'name': 'kraken'}
+                        ).dict()
+                    )
+            except AssertionError:
+                log.error(f'Order cancel was not successful')
```
goodboy, line 628: hmm could you also maybe send a `BrokerdError` msg in this case?

piker/brokers/kraken.py
```diff
@@ -271,18 +454,304 @@ async def bars(
...
+@tractor.context
+async def trades_dialogue(
+    ctx: tractor.Context,
+    loglevel: str = None,
+) -> AsyncIterator[Dict[str, Any]]:
+
+    # XXX: required to propagate ``tractor`` loglevel to piker logging
+    get_console_log(loglevel or tractor.current_actor().loglevel)
+
+    # Generate
```
goodboy, line 643: wut?

piker/brokers/kraken.py
```diff
@@ -271,18 +454,304 @@ async def bars(
...
+
+        # unsub from all pairs on teardown
+        await ws.send_msg({
+            'event': 'unsubscribe',
+            'subscription': ['ownTrades'],
+        })
+
+        # XXX: do we need to ack the unsub?
+        # await ws.recv_msg()
+
```
goodboy, line 671: isn't this too many newlines?

piker/brokers/kraken.py
```diff
@@ -271,18 +454,304 @@ async def bars(
...
+                            broker_time=float(trade.broker_time)
+                        )
+                        
+                        await ems_stream.send(fill_msg.dict())
+
+
+async def stream_messages(
+    ws: NoBsWs,
+):
+    '''
```
goodboy, line 748: This is the doc str style i'm kinda fond of. It's a little less noisy imo but willing to be challenged on it of course.

piker/brokers/kraken.py
```diff
@@ -320,39 +789,93 @@ async def stream_messages(ws):
...
+        #     print(msg)
+
+        else:
+            print(f'UNHANDLED MSG: {msg}')
+            yield msg
+
+
+async def process_trade_msgs(
+    ws: NoBsWs,
+):
```
goodboy, line 838: Hmm not that i know the answer right away (i struggle with this as well, think it's `AsyncGenerator`?) but we should probably add return type annots here.

piker/brokers/kraken.py
```diff
@@ -320,39 +789,93 @@ async def stream_messages(ws):
...
+    Parse and pack data feed messages.
+
+    '''
+    sequence_counter = 0
+    async for msg in stream_messages(ws):
+
+        try:
+            # check that we are on the ownTrades stream and that msgs are 
+            # arriving in sequence with kraken
+            assert msg[1] == 'ownTrades'
```
goodboy, line 849: maybe link to api docs on the indexing details here.

piker/data/_web_bs.py
goodboy: you could pass all (future) data like this as a `fixture_kwargs: dict` and then maybe just call it with `self.fixture(self, **self.fixture_kwargs)` and i think as long as the fixture doesn't declare it as a default kwarg it should just work?

goodboy (changes requested):
@iamzoltan lookin gud.

I manually tested any besides comments about cancel latency this is feeling correct.
I obviously left a lot of stylistic comments for change but this is shaping up really nicely>

In particular i really like the pedantic error checking where possible 👍🏼

piker/brokers/kraken.py
```diff
@@ -14,13 +14,13 @@
...
-"""
+'''
 Kraken backend.
 
-"""
+'''
 from contextlib import asynccontextmanager
 from dataclasses import asdict, field
-from typing import List, Dict, Any, Tuple, Optional
+from typing import Dict, List, Tuple, Any, Optional, AsyncIterator
```
goodboy, line 23: Ok so I wanted these type annots changed to 3.9+ style but i actually already did this in history in #247 😂 so maybe we can get the conflicts merged in via git then instead of you doing it manually.

piker/brokers/kraken.py
goodboy: Ok so we're not going to try and tackle this i guess?

If we end up moving to `cryptofeed` anyway this might all be in vain. so yeah not gonna expect this any more.

piker/brokers/kraken.py
iamzoltan: I have not and will get on it

piker/brokers/kraken.py
goodboy: Ok let's postpone and make an issue for this for now.

we can check how `cryptofeed` does this and maybe just delegate to them?

piker/brokers/kraken.py
```diff
@@ -162,6 +234,134 @@ async def _public(
...
+        # https://docs.kraken.com/rest/#operation/getTradeHistory
+        while True:
+            resp = await self.endpoint('TradesHistory', data)
+            # grab the first 50 trades
+            if data['ofs'] == 0:
+                trades = resp['result']['trades']
+            # load the next 50 trades using dict constructor
+            # for speed
+            elif data['ofs'] == 50:
+                trades = dict(trades, **resp['result']['trades'])
```
goodboy, line 285: consider this new 3.9+ merge syntax for `dict`:
https://docs.python.org/3/whatsnew/3.9.html#dictionary-merge-update-operators

piker/brokers/kraken.py
goodboy: `cryptofeed` point's of interest:
- [trades parsing mixin method](https://github.com/bmoscon/cryptofeed/blob/master/cryptofeed/exchanges/mixins/kraken_rest.py#L112), which calls
- [`._historical_trades()](https://github.com/bmoscon/cryptofeed/blob/c37f9c4fc19cafa877eda8ee5ceb2c1c3c17a830/cryptofeed/exchanges/mixins/kraken_rest.py#L127) which has some very [interesting edge case handling](https://github.com/bmoscon/cryptofeed/blob/c37f9c4fc19cafa877eda8ee5ceb2c1c3c17a830/cryptofeed/exchanges/mixins/kraken_rest.py#L133) which seems to use some kinda `&since=` http params flag? I dunno.

piker/brokers/kraken.py
goodboy: instead of all this key error stuff use `positions.setdefault(trade['pair'], sign * float(trade['cost']))`

piker/brokers/kraken.py
```diff
@@ -162,6 +234,134 @@ async def _public(
...
+
+        # positions
+        # TODO: Make sure to add option to include fees in positions calc
+        for trade in trades.values():
+            sign = -1 if trade['type'] == 'sell' else 1
+            # This catch is for populating the dict with new values
+            # as the plus assigment will fail if there no value
+            # tied to the key
+            try:
+                positions[trade['pair']] += sign * float(trade['cost'])
```
goodboy, line 309:
```python
    for trade in trades.values():
            sign = -1 if trade['type'] == 'sell' else 1
            # This catch is for populating the dict with new values
            # as the plus assigment will fail if there no value
            # tied to the key
           pair = trade['pair']
           positions.setdefault(pair, 0) += sign * float(trade['cost'])
           vols.setdefault(pair, 0) += sign * float(trade['vol'])
```

goodboy: even better getting rid of the lower loop

```python
for trade in trades.values():
    sign = -1 if trade['type'] == 'sell' else 1
    # This catch is for populating the dict with new values
    # as the plus assigment will fail if there no value
    # tied to the key
    pair = trade['pair']
    positions.setdefault(pair, 0) += sign * float(trade['cost']) / float(trade['vol'])
```

goodboy: This should work, *i think*:

```python
pairs: dict[str, float] = {}
vols: dict[str, float] = {}

for trade in trades.values():
    sign = -1 if trade['type'] == 'sell' else 1
    # This catch is for populating the dict with new values
    # as the plus assigment will fail if there no value
    # tied to the key
    pair = trade['pair']
    vol = float(trade['vol']))
    vols.setdefault(pair, 0) += vol
    positions.setdefault(pair, 0) += (sign * float(trade['cost']) * vol) / vols[pair]
```

piker/brokers/kraken.py
goodboy: See above pretty sure we can drop this loop.

piker/brokers/kraken.py
iamzoltan: will remove

piker/data/_web_bs.py
iamzoltan: will be address in a new PR, see here #290

piker/brokers/kraken.py
iamzoltan: #290

piker/brokers/kraken.py
goodboy: ah sorry that was the wrong endpoint.

You actually probably want this [`.trade_history()` method](https://github.com/bmoscon/cryptofeed/blob/c37f9c4fc19cafa877eda8ee5ceb2c1c3c17a830/cryptofeed/exchanges/mixins/kraken_rest.py#L230) which is private and uses the same actual web ep.

goodboy: yeah [i don't think they compute pps whatsoever](https://github.com/bmoscon/cryptofeed/search?q=position+kraken) in `cryptofeed` so we're going to need this implementation either way 🏄🏼

piker/brokers/kraken.py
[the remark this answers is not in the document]
goodboy: As mentioned in chat stick all the pp calcing stuff in this routine and keep the method `get_positions()` to be something like `get_trades()` and just have it pull from the API. This will also make is easier to swap in similar `cryptofeed` apis later if need be.

guilledk (approved):

piker/brokers/kraken.py
iamzoltan: got a working version going, there is a second loop, but that is simply to package the positions in the msg format.

piker/brokers/kraken.py
goodboy: lol just tried it. it's worse case: just halts and blocks 😂

piker/brokers/kraken.py
```diff
@@ -271,18 +446,333 @@ async def bars(
...
+    trades: dict
+) -> list[Any]:
+    positions: dict[str, float] = {}
+    vols: dict[str, float] = {}
+    costs: dict[str, float] = {}
+    position_msgs: list[Any] = []
+
+    for trade in trades.values():
+        sign = -1 if trade['type'] == 'sell' else 1
+        # This catch is for populating the dict with new values
```
goodboy, line 477: drop this legacy comment

piker/brokers/kraken.py
```diff
@@ -162,6 +234,109 @@ async def _public(
...
+        data: Dict[str, Any]
+    ) -> Dict[str, Any]:
+        uri_path = f'/0/private/{method}'
+        data['nonce'] = str(int(1000*time.time()))
+        return await self._private(method, data, uri_path)
+
+    async def get_trades(
+        self,
+        data: Dict[str, Any] = {}
+    ) -> Dict[str, Any]:
```
goodboy, line 271: change to `dict`

goodboy: or wait did we just bail on all that?

piker/brokers/kraken.py
```diff
@@ -162,6 +234,109 @@ async def _public(
...
+            elif resp['result']['trades'] == {}:
+                count = resp['result']['count']
+                break
+            # update existing dict if num trades exceeds 100
+            else: 
+                trades.update(resp['result']['trades'])
+            # increment the offset counter
+            data['ofs'] += 50
+            # To avoid exceeding API rate limit in case of a lot of trades
+            await trio.sleep(1)
```
goodboy, line 295: you should define this throttle value somewhere with a full name i'm thinking, and link to docs.

piker/brokers/kraken.py
```diff
@@ -162,6 +234,109 @@ async def _public(
...
+        return trades
+
+    async def submit_limit(
+        self,
+        oid: str,
+        symbol: str,
+        price: float,
+        action: str,
+        size: float,
+#        account: str,
```
goodboy, line 309: stale?

piker/brokers/kraken.py
```diff
@@ -271,18 +446,333 @@ async def bars(
...
+    for trade in trades.values():
+        sign = -1 if trade['type'] == 'sell' else 1
+        # This catch is for populating the dict with new values
+        # as the plus assigment will fail if there no value
+        # tied to the key
+        pair = trade['pair']
+        vol = float(trade['vol'])
+        # This is for the initial addition of a pair so the 
+        # += operation does not fail.
+        vols[pair] = vols.setdefault(pair, 0)
```
goodboy, line 484: Sorry why are you doing what seems to be a needless assignment here?

goodboy: In fact the more terse way to do this would be to avoid the `+=` syntax and drop the half the lines:

```python
vols[pair] = vols.setdefault(pair, 0) + sign * vol
```

piker/brokers/kraken.py
```diff
@@ -271,18 +446,333 @@ async def bars(
...
+        costs[pair] = costs.setdefault(pair, 0)
+        positions[pair] = positions.setdefault(pair, 0)
+        vols[pair] += sign * vol
+        costs[pair] += sign * float(trade['cost'])
+        if vols[pair] != 0:
```
goodboy, line 489: you could probably simplify this to:

```python
positions[pair] = costs[pair] / vols[pair] if vols[pair] else 0
```

piker/brokers/kraken.py
```diff
@@ -271,18 +446,333 @@ async def bars(
...
+            positions[pair] = costs[pair] / vols[pair]
+        else:
+            positions[pair] = 0
+
+    for ticker, pos in positions.items():
+        norm_sym = normalize_symbol(ticker)
+        vol = float(vols[ticker])
+        if vol != 0:
```
goodboy, line 497: i would consider inverting this:

```python
if not vol:
    continue
```

goodboy (changes requested):
@iamzoltan very very close and good work!

Just a few more changes I'd recommend to make things a bit terser and you still need to fix the case there the `[kraken]` config section is missing.

piker/brokers/kraken.py
goodboy: mabye don't worry about the type annots since i think i already have a patch with them all

piker/brokers/kraken.py
goodboy: heh, actually come to think of it, if you *must* do an assignment then you don't really need `.setdefault()`; just use `vols.get(pair, 0)` instead in the first term.

piker/brokers/kraken.py
goodboy: actually, i think you should drop this and handle whatever throttle error they throw instead.
as a user's trade log get's large this is going to cause super slow startup.

piker/brokers/kraken.py
iamzoltan: should I do something about it in this PR?

piker/brokers/kraken.py
iamzoltan: yea we did bail on it for now

piker/brokers/kraken.py
iamzoltan: yea I could drop it, and tbh, I dont think you would run into the errors unless you have a long list of trades. The docs have a timer point system that throttles the API calls. I just had it there because my logic was bugged at the beginning and I was calling too much. We could actually have the sleep value at the limit of their throttling, thoughts?

piker/brokers/kraken.py
iamzoltan: the `account` argument? ye seems like it

piker/brokers/kraken.py
iamzoltan: yessir

piker/brokers/kraken.py
iamzoltan: yea I like the `.get(pair, 0)` approach better. will get on it

piker/brokers/kraken.py
iamzoltan: this is now addressed in #290

piker/brokers/kraken.py
iamzoltan: hang is fixed, but need to figure out how to boot up paper

piker/brokers/kraken.py
```diff
@@ -764,6 +764,23 @@ async def subscribe(ws: wsproto.WSConnection, token: str):
...
 
+                        filled_msg = BrokerdStatus(
+                            reqid=trade.reqid,
+                            time_ns=time.time_ns(),
+
+                            account='kraken.spot',
+                            status='filled',
+                            filled=float(trade.size),
+                            reason='Order filled by kraken',
+                            # remaining='' # TODO: not sure what to do here.
```
goodboy, line 775: yeah it's because you haven't seen the general case.
basically a fill can *use up* (aka subtract from) the current order's "size".

So say you put in an order for 1000 of something but only 400 get's filled, then shortly later another 600; that would terminate the order's lifetime - it would be completely. However multiple fills can be used to "consume" up the order. More technically, and order can be completed over multiple *clearing prices*.

goodboy: made https://github.com/pikers/piker/issues/296 as follow up.

goodboy:
@iamzoltan yeah so i think just because this PR is so long in the tooth we should just move improvements to new patch sets so that i ane others can start adding more stuff to this backend (like hft history retrieval). It's also nicer to be able to avoid merging conflicts in smaller chunks going forward.

Frankly there's still a ton of stuff I'd normally just keep rejecting (mostly style wise) but I think for momentum's sake it might be better to just get this in and i can take care of some pedantic stuff in some later patches :)

piker/brokers/kraken.py
goodboy: Yeah so the main thing you're missing here is pretty simple:

- you've got your `Trade(BaseModel)` type that is used to parse the details from kraken
- you've also got a message you **shouldn't be sending** which is the first one with the `status=executed`. The reason you didn't catch this is bc [we're only logging that message](https://github.com/pikers/piker/blob/master/piker/clearing/_ems.py#L705) instead of making sure it's in the correct set; we should make a bug report for this.
- you've got 2 msgs you need to send to the ems:
  - a `BrokerdStatus` with status 'filled' which indicates that (some of) the order was filled and there is [logic in the ems's relay loop](https://github.com/pikers/piker/blob/master/piker/clearing/_ems.py#L689) which checks the `.remaining` field on that message. [If the remaining field is `bool` null (`0` or `None`) then the ems takes care of sending the `broker_executed` msg to the client side](https://github.com/pikers/piker/blob/master/piker/clearing/_ems.py#L693). This message doesn't have to declare the fill/execution details if the backend does not want to but *it can* in the following additional message
  - a `BrokerdFill` message *can* be sent if you expect multiple fills to eventually fulfill the *execution* (i.e. all clears of) the logical *order*. These msgs [are also relayed to the client](https://github.com/pikers/piker/blob/master/piker/clearing/_ems.py#L712) which allows it to draw multiple annotations on the chart (triangles) that should all the individual fills that eventually satisfy the order's execution.

TL;DR:
- don't send the first msg, it's noto valid but we need to actually enforce that
- the `filled` status should decrement it's `.remaining` on every `Trade` you get from the kraken end
- `BrokerdFill` msgs should be sent 1-to-1 for every `Trade` msg you get from kraken so that the UI can show the multiple clears on chart.

goodboy: made https://github.com/pikers/piker/issues/297 to address why you originally were sending the wrong `.status`.

goodboy (approved):
All follow ups are in issues now so just gonna land this puppy.

piker/brokers/kraken.py
iamzoltan: thanks, this sounds great. it is also much more clear now. I like the idea of enforcing not making the mistake I did.

PR #218, opened 2021-09-05
Paper pp tracking

goodboy:
Factors out some commits from #205 in an effort to make that patch set smaller for review.

Main points of interest:
- pp tracking with line and marker for paper engine positions (what we expect new users to play with)
- re-org `._graphics` subpkg into top level `.ui` subpkg


piker/ui/order_mode.py
```diff
@@ -41,11 +41,30 @@
...
+
+    '''
     symbol: Symbol
     size: float
-    avg_price: float
+    avg_price: float  # TODO: contextual pricing
+    fills: Dict[str, Any] = {}
+
+
+class OrderDialog(BaseModel):
```
goodboy, line 53: Ahh right this was the other main client-side addition:  a type to keep track of the order messages for easier analysis by the order mode code.

PR #223, opened 2021-09-09
`brokerd`, `emsd` and UI multi-account per broker, order mode support 

goodboy:
Makes the *account* selection field in the order mode sidepane actually work fully on `ib`.

There's still a few more design things to change:
- [x] we need an `Allocator` *per account* which should be read into the status UI on user switches
- [x] the order mode sidepane needs to distinguish status elements depending on what account is selected; the health bar displays the pp for the currently selected account
   - do we want to have this update automatically on mouse overs?
- [x] show account names on order and pp lines
---
deferred to #224 :
 - [ ] add a config option to let the user choose whether to always show all account pps or just the current?
 - [ ] would-be-nice, a little dot next to account names in the selection list that have open pps


piker/ui/order_mode.py
```diff
@@ -849,4 +860,6 @@ async def process_trades_and_update_ui(
                 arrow_index=get_index(details['broker_time']),
             )
 
-            tracker.live_pp.fills.append(msg)
+            # TODO: how should we look this up?
```
goodboy, line 863: This is an outstanding question.

2 ways to do it:
- create an `oid` -> tracker map here
- add an `account` field to `BrokerdFill` msgs

config/brokers.toml
```diff
@@ -14,12 +14,14 @@ private_key = ""
...
-paper = ""
+ports.gw = 4002
+ports.tws = 7497
+ports.order = ["gw", "tws",]
 
-[ib.ports]
-gw = 4002
-tws = 7497
-order = [ "gw", "tws",]
+accounts.margin = "X0000000"
```
goodboy, line 21: So basically all providers can define a table like this now.

We might tweak the schema a bit but i think this is simple enough to start.

piker/brokers/ib.py
```diff
@@ -220,15 +219,18 @@ class Client:
     Note: this client requires running inside an ``asyncio`` loop.
 
     """
+    _contracts: dict[str, Contract] = {}
```
goodboy, line 222: class scoped and thus global from now on.

piker/brokers/ib.py
```diff
@@ -1314,28 +1369,57 @@ def pack_position(pos: Position) -> dict[str, Any]:
...
 
         if action in {'buy', 'sell'}:
             # validate
             order = BrokerdOrder(**request_msg)
 
             # call our client api to submit the order
-            reqid = await _trio_run_client_method(
-
-                method='submit_limit',
+            reqid = client.submit_limit(
```
goodboy, line 1416: turned out we can just call these synchronously which is kinda handy.

PR #225, opened 2021-09-12
Ugh, positions relay hotfix

goodboy:
Remnant from #223.

Must have run into some confusion with data structures in `brokerd` vs.
`emsd`. This fixes the ems `relay.positions` state tracking to be
composed maps, vs. messages from `brokerd` should just be a sequence.

### DIFF
--- a/piker/brokers/ib.py
+++ b/piker/brokers/ib.py
@@ -1465,7 +1465,7 @@ async def trades_dialogue(
     global _client_cache
 
     # deliver positions to subscriber before anything else
-    all_positions = {}
+    all_positions = []
 
     clients: list[tuple[Client, trio.MemoryReceiveChannel]] = []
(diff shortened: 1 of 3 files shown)


piker/brokers/ib.py
```diff
@@ -1480,9 +1483,8 @@ async def trades_dialogue(
         for client in _client_cache.values():
             for pos in client.positions():
                 msg = pack_position(pos)
-                all_positions.setdefault(
-                    msg.symbol, []
-                ).append(msg.dict())
+                msg.account = accounts_def.inverse[msg.account]
```
goodboy, line 1486: Added in this: sends the user defined (in `brokers.toml`) account *name* now instead of the actual account number/value over the wire.

This probably makes more sense anyway once a more distributed setup is going on (`brokerd` on diff hosts then GUI apps).

Issue #224, opened 2021-09-12
Account "liveness", has pp?,  UX


goodboy:
In #223 we land multi-account support per `brokerd` an now support position display and orders over multiple broker accounts on the same order mode for a chart.

Some lingering UX questions and options i had thought of include how to display the state of which accounts both are *loaded and live* - ready for order submissiong, as well as which of those accounts have existing positions per symbol and can we show this succinctly both in the search view and the account selection view.

Some ideas:
- always load all accounts defined (per broker) in `brokers.toml` **and** show an `x` graphic beside accounts entries which weren't loaded/connected to indicate that switching to these will be a no-op
- **or** only load accounts that can be currently used/connected for the symbol and add a *refresh* control that would allow (re)loading accounts incrementally (say after starting some remote ``brokerd``)
- once you have a loaded account, show a small `pp` or dot graphic beside each account that has an existing position (possibly in green/red to indicate pnl polarity)

goodboy:
In #226 there's 2 un-done bullets that can be used as resources as well.

The current state after that lands is that we have long/short icons (with the same graphic as our level marker for pps) and we're only loading accounts that the `brokerd` reports to us on connection to the EMS.

goodboy:
Another thing we should add is the new pp marker icons to entries in our custom [search view](https://github.com/pikers/piker/blob/master/piker/ui/_search.py#L81). 

[this SO](https://stackoverflow.com/a/38308129) shows that you just use `QStandardItem.setIcon()`. we'll prolly want to add a `pp` column to the headers as well.

I've been thinking we might actually want to populate a section as well with pps for each `brokerd` and just have the feeds be unloaded until a user selects one?

Issue #124, opened 2020-11-24
Caching past datums for fast render: `QPixMap`, `QPainterPath`, `QBackingStore` or OpenGL?


goodboy:
In trying to squeeze performance for #109 I stumbled upon many resources for toying with avoidance of graphics re-drawing/rendering and have now a slew of place to get started if we decide to dig further on this optimization front in the `Qt` types and apis.

This also pertains to ongoing work discussed in #109 which has since moved to rendering OHLC bars using `QPainterPath` much like `pyqtgraph`'s internal usage for line plots which are very performant even for moderately sized data sets.

As some preface please note the following:
- `pyqtgraph` uses the [Graphics View](https://doc.qt.io/qt-5/graphicsview.html) framework internally for which it overrides and extends many types, in particular the [`GraphicsItem`](https://doc.qt.io/qt-5/qgraphicsitem.html#details) which is used as the base for almost everything
- a [primer on modern usage/discussion](https://www.qt.io/blog/2017/01/19/should-you-be-using-qgraphicsview) of `QGraphicsView` (make sure to read follow up comments)
- knowledge of the [`QPainter`](https://doc.qt.io/qt-5/qpainter.html) api is a must
- take a cursory view of the [2d drawing options in `Qt`](https://doc.qt.io/qt-5/qtgui-index.html#2d-graphics)

#### `QPainterPath` (what we're using)
- `pyqtgraph` uses this internally with a fancy binary-format array generation routine, [`pg.functions.arrayToQPath()`](https://github.com/pyqtgraph/pyqtgraph/blob/master/pyqtgraph/functions.py#L1441), which [streams to the path](https://github.com/pyqtgraph/pyqtgraph/blob/master/pyqtgraph/functions.py#L1552) magically and is quite fast
- we've adopted this as the implementation for our internal `BarsItems` graphics and it seems to be scaling well (at least way better then any other easily testable platform) with renders of 15-20k bars with ~ 3ms latency.
- we've `numba`-ized the binary format array generation which can likely be extended if `arrayToQPath()` can be split up into it's `numpy` only and `Qt` only parts.
- appends are supported out of the box using `.addPath()` (which we're also using) which avoids having to regenerate path segments previously drawn to screen
- use of `QPicture` and `QPainter.drawPicture()` seems to slow this down oddly
- the [painter paths example](https://doc.qt.io/qt-5/qtwidgets-painting-painterpaths-example.html) is likely relevant reading for newcomers to this api

[...]

PR #226, opened 2021-09-13
Account select icons

goodboy:
A first draft towards #224.

Tried a few things (with some minor success) but still not happy with it.

---
The list of problems on [adding icons to `QComboBox`](https://doc.qt.io/qt-5/qcombobox.html#setItemIcon):
- right now we're using the [built-in icons set](https://www.pythonguis.com/faq/built-in-qicons-pyqt/) but there's also ways to get [linux-specific icons](https://specifications.freedesktop.org/icon-naming-spec/latest/ar01s04.html) (which seem to be loaded [like this](https://stackoverflow.com/questions/7253976/how-do-i-add-an-icon-to-qcombobox-in-qt)) as is mentioned in this blog post
  - here's the hot [SO tips](https://stackoverflow.com/questions/52358982/how-to-use-qstylestandardicon-standardpixmap-with-qstylestandardpixmap) on coverting std icons to `QIcon`
- trying to get icons to show up on the RHS (the most infuriating part so far):
    - [x] [customizing the `.paint()` on the delegate](https://stackoverflow.com/questions/39941183/pyqt-add-icon-to-right-side-of-qlistwidget-item) worked for **the view items only** (**not** the current item label) but it seems like we should be able to set this once up front no? (maybe somehow setting a [`QStyleOptionComboBox`](https://doc.qt.io/qt-5/qstyleoptioncombobox.html), but no idea where to do that...)
    - [ ] to try, this SO on [hiding the icon in the current item label](https://stackoverflow.com/questions/45546155/hide-icon-from-the-label-of-qcombobox) which is hacky as heck but seems to have give some insight?
- trying to get the icons themselves to match our color scheme / using custom graphics
 
    - [x] trying to fill the built-in icons using [a pixmap mask hack](https://stackoverflow.com/questions/13350631/simple-color-fill-qicons-in-qt)
    - [ ] it may be possible to generate pixmap icons from polygons, not sure if we need to add a [`.drawIcon()`](https://doc.qt.io/qt-5/qicon.html#making-classes-that-use-qicon) method or not (to what though?) but this might play with the `QPixMap` section we have in #124.
      - tried this using what is the [std apis](https://stackoverflow.com/questions/17888795/how-to-use-qpainter-on-qpixmap) and couldn't get it to work or look good at all. left a `TODO` in the icon generation code..
  
  
 ---
 Other `Qt` notes from this tinkering:
 -  it might be possible to specify view item heights [without a custom delegate](https://stackoverflow.com/questions/13308341/qcombobox-abstractitemviewitem?rq=1)

### DIFF
--- a/piker/clearing/_paper_engine.py
+++ b/piker/clearing/_paper_engine.py
@@ -463,7 +463,7 @@ async def trades_dialogue(
         # TODO: load paper positions per broker from .toml config file
         # and pass as symbol to position data mapping: ``dict[str, dict]``
         # await ctx.started(all_positions)
-        await ctx.started({})
+        await ctx.started(({}, {'paper',}))
 
         async with (
             ctx.open_stream() as ems_stream,
--- a/piker/brokers/ib.py
+++ b/piker/brokers/ib.py
@@ -1469,6 +1469,7 @@ async def trades_dialogue(
 
     # deliver positions to subscriber before anything else
     all_positions = []
+    accounts = set()
 
     clients: list[tuple[Client, trio.MemoryReceiveChannel]] = []
     for account, client in _accounts2clients.items():
@@ -1484,9 +1485,10 @@ async def trades_dialogue(
             for pos in client.positions():
                 msg = pack_position(pos)
                 msg.account = accounts_def.inverse[msg.account]
+                accounts.add(msg.account)
                 all_positions.append(msg.dict())
 
-    await ctx.started(all_positions)
+    await ctx.started((all_positions, accounts))
 
     async with (
         ctx.open_stream() as ems_stream,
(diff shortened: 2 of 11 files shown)


goodboy:
Bleh, yeah I gave up on the RHS icons; it's just too much hackery and not even sure it'll hold up with Qt6.

I think the LHS is fine for now since it brings your attention immediately to which accounts have live pps.

piker/clearing/_paper_engine.py
```diff
@@ -463,7 +463,7 @@ async def trades_dialogue(
         # TODO: load paper positions per broker from .toml config file
         # and pass as symbol to position data mapping: ``dict[str, dict]``
         # await ctx.started(all_positions)
-        await ctx.started({})
+        await ctx.started(({}, {'paper',}))
```
goodboy, line 466: My last question is maybe this should just be an explicit map?

```python
{'positions': {}, 'accounts': {'paper',}}
```

kinda thing?

PR #230, opened 2021-09-21
Super basic brokerd status

goodboy:
Gives us a little label on the order mode settings pane that shows **super** (for now) basic info about the quotes feed providing `brokerd` actor.

Basically looks like this:

screenshot-2021-09-21_15-54-15

Eventually we want to have some controls here for restarting feeds and full actors as needed as well as for *moving* a stream source to another remote actor / `pikerd` cluster.

This is just a super basic draft label with initial info to get things going.

Also included is some new `puterize()` func logic which let's you input values with the trader-slang suffixes `{'k', 'M', 'B'}`, basically the opposite of our `piker.calc.humanize()`.

### DIFF
--- a/piker/calc.py
+++ b/piker/calc.py
@@ -20,10 +20,17 @@ Handy financial calculations.
 import math
 import itertools
 
+from bidict import bidict
+
+
+_mag2suffix = bidict({3: 'k', 6: 'M', 9: 'B'})
+
 
 def humanize(
+
     number: float,
     digits: int = 1
+
 ) -> str:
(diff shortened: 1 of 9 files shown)


piker/calc.py
```diff
@@ -20,10 +20,17 @@
 import math
 import itertools
 
+from bidict import bidict
+
+
+_mag2suffix = bidict({3: 'k', 6: 'M', 9: 'B'})
```
goodboy, line 26: are there any other ones i'm not thinking of?

`m` seems like a worse version of `k`.

Issue #216, opened 2021-09-01
Expose FSP streams as `Feed`s


goodboy:
With #212 comes some nifty features like task broadcasting (from `tractor`) and quote stream *pausability* for keeping tasks switches in Qt apps as low as possible for performance.

I'm thinking taking the [`piker.data.feed.Feed`](https://github.com/pikers/piker/blob/master/piker/data/feed.py#L339) api to our fsp subsystem would be a nice abstraction and would extend these features to sub-plots which are currently not being throttled the same way as OHLC charts.

I *think* the only thing we need to really add is an endpoint argument that the underlying portal will call, right now that's by default [`data.feed.attach_feed_bus()`](https://github.com/pikers/piker/blob/master/piker/data/feed.py#L474). We'll likely want to generalize the `Feed` instance var names (like [the portal ref](https://github.com/pikers/piker/blob/master/piker/data/feed.py#L353)) and further add uids for each fsp and/or the ability to attach to existing symbol-fsp pairs for stream and signal sharing.

PR #231, opened 2021-09-21
Vlm plotz

goodboy:
WIP built-in volume sub-plot curve with a bunch of fancy styling additions.

Of note:
- we can't use a "filled curve" bc it kills performance during high trade rates - seems to be some kinda Qt issue that we gotta dig into, i already spoke to `pyqtgraph` core about it
- made all "current" step graphics be *highlighted* in some way vs. the preceding history; i personally think this looks really slick and makes focusing on what is current vs. past very simple
- there's still some history here pertaining to very basic work toward #216 which will likely get factored out into another PR

Still TODO:
- [x] $vlm support via a typical price function on the price (per bar) *times* the vlm each step
  - example implementation using `ohl3` [on tinaview](https://www.tradingview.com/script/O3xqRCCi-Total-Dollar-Volume/)
  - related DVL def [from investopedia](https://www.investopedia.com/terms/d/dollar-volume-liquidity.asp)
- ~[ ] support for flipping the "units" from $ to shares in the sidepane~ ended up going with dual axes instead 😎 

Possibly more to come.

### DIFF
--- a/notes_to_self.rst
+++ b/notes_to_self.rst
@@ -1,0 +1,28 @@
+Notes to self
+=============
+chicken scratch we shan't forget, consider this staging
+for actual feature issues on wtv git wrapper-provider we're
+using (no we shan't stick with GH long term likely).
+
+
+cool chart features
+-------------------
+- allow right-click to spawn shell with current in view
+  data passed to the new process via ``msgpack-numpy``.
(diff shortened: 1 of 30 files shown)


goodboy:
Was brought in as a series of other PRs.

PR #235, opened 2021-10-29
Ib client scanning

goodboy:
Adds support for using the new `tractor.to_asyncio.open_channel_from()` cross-async-loop api and vastly improves API endpoint scanning and cacheing for the ib backend.

Some other fixes and improvements:
- only return from a `Client.get_quote()` call once a "last price" has actually been filled in (previously due to asynchrony this would sometimes be `nan`..)
- add a mostly working data feed reset hack script which works around hitting throttle limits on history queries

### DIFF
--- a/piker/_daemon.py
+++ b/piker/_daemon.py
@@ -47,7 +47,7 @@ _root_modules = [
 
 class Services(BaseModel):
 
-    actor_n: tractor._trionics.ActorNursery
+    actor_n: tractor._supervise.ActorNursery
     service_n: trio.Nursery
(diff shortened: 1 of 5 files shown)


goodboy:
Oh right we need to move our tractor pin onto `infect_asyncio` latest again to make this work.

goodboy:
Ok this basically **must** coincide with a `piker_pin` branch update on `tractor`. So I'm going to push that change then merge this right away instead of worrying about merges into that branch (i.e. did a rebase and force push - what could go wrong 😂 ).

PR #237, opened 2021-11-03
Order pane fixes

goodboy:
Miscellaneous fixes for the order mode pane UI.

Summary:
- fix a couple edge cases for discrete tradables (eg. stocks) where on a final exit of a single unit rounding logic caused a zero size and another case where a short sized negative `int` would generally mess up exit (buying to cover) sizing. 
- fixed some pnl label task startup/resume logic and label re-sizing issues that stemmed from increasing the number of shown digits on magnitude steps (eg. 0-9 -> 10-99 would cause label resizes which were quite obnoxious).
- drop use of `Enum` from the allocator (can't seem to make any sense of how `pydantic` expects you to use it..)
- expect a `tuple` of accounts from the EMS (instead of a set) in order to both preserve the user's config order of names as well as not break `msgpack` 😂.

### DIFF
--- a/piker/clearing/_allocate.py
+++ b/piker/clearing/_allocate.py
@@ -87,13 +87,21 @@ class Allocator(BaseModel):
 
     symbol: Symbol
     account: Optional[str] = 'paper'
-    size_unit: SizeUnit = 'currency'
+    # TODO: for enums this clearly doesn't fucking work, you can't set
+    # a default at startup by passing in a `dict` but yet you can set
+    # that value through assignment..for wtv cucked reason.. honestly, pure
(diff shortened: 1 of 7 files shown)


iamzoltan:
looks better! the added comments help provide some clarity and I like the cleaned up naming. Will pull and test it ooot!

goodboy:
> Will pull and test it ooot!

just don't pull too hard.

goodboy:
@iamzoltan ok master should be fixed once we land #241 now and this has been rebased onto that.

iamzoltan:
tested it out, works now! thanks. good to merge imo

PR #239, opened 2021-11-03
Draft tina install section

goodboy:
@wattygetlood please review and add to as desired 😂 

I think this is a decent start from what I remember helping with.


guilledk (approved):
How do you know all this tina stuff lol

goodboy:
I guess we can always fix it up better later 😂

PR #240, opened 2021-11-03
Fast step curve

goodboy:
This adds a so called "step curve" graphic that is drawn as sequence of line segments per discrete datum in the domain. It can be thought of what a histogram typically looks like except we don't draw "separator lines" (vertical segments that touch the zero line) between each datum thus giving the appearance of a continuous curve where for each sample in the x, there is a flat segment in the y spanning that sample index in the x.

`pyqtgraph` actually already has this feature [inside the `PlotCurveItem` type](https://github.com/pyqtgraph/pyqtgraph/blob/8436457cc38b66d060a0d5afd7c92fb0a2cca4cb/pyqtgraph/graphicsItems/PlotCurveItem.py#L537) but it's somewhat hard to configure correctly to our sampling style (OHLC typically) and is **not** optimized for real-time update like our rewrite of `PlotCurveItem`. Further the `fillLevel` support is horrendously slow is used with `stepMode` as is demonstrated (and hopefully solved) in https://github.com/pyqtgraph/pyqtgraph/pull/2032.

I actually personally prefer the style now of this non-filled version anyway but we may want to support it if it's improved in `pyqtgraph` core.

Some additional niceties include highlighting the most recent curve-step-segment to distinguish it from history data. It's much easier to see what is the live data step, even from a far out zoom.

Also note, this PR does not include any actual usage of the new graphic 😂

That is all coming in #231.

### DIFF
--- a/piker/ui/_curve.py
+++ b/piker/ui/_curve.py
@@ -137,6 +137,9 @@ class FastAppendCurve(pg.PlotCurveItem):
         # print(f"xrange: {self._xrange}")
         istart, istop = self._xrange
 
+        # compute the length diffs between the first/last index entry in
+        # the input data and the last indexes we have on record from the
+        # last time we updated the curve index.
         prepend_length = istart - x[0]
         append_length = x[-1] - istop
 
@@ -149,7 +152,7 @@ class FastAppendCurve(pg.PlotCurveItem):
             # by default we only pull data up to the last (current) index
             x_out, y_out = x[:-1], y[:-1]
 
-        if self.path is None or prepend_length:
+        if self.path is None or prepend_length > 0:
             self.path = pg.functions.arrayToQPath(
                 x_out,
                 y_out,
@@ -177,7 +180,7 @@ class FastAppendCurve(pg.PlotCurveItem):
         #     # self.path.moveTo(new_x[0], new_y[0])
         #     self.path.connectPath(old_path)
 
-        elif append_length:
+        elif append_length > 0:
             if self._step_mode:
                 new_x, new_y = step_path_arrays_from_1d(
                     x[-append_length - 2:-1],
--- a/piker/ui/_curve.py
+++ b/piker/ui/_curve.py
@@ -22,9 +22,87 @@ from typing import Tuple
 
 import numpy as np
 import pyqtgraph as pg
-from PyQt5 import QtCore, QtGui, QtWidgets
+from PyQt5 import QtGui, QtWidgets
+from PyQt5.QtCore import (
+    QLineF,
+    QSizeF,
+    QRectF,
+    QPointF,
+)
 
 from .._profile import pg_profile_enabled
+from ._style import hcolor
+
+
+def step_path_arrays_from_1d(
+    x: np.ndarray,
+    y: np.ndarray,
+
+) -> (np.ndarray, np.ndarray):
+    '''Generate a "step mode" curve aligned with OHLC style bars
+    such that each segment spans each bar (aka "centered" style).
+
+    '''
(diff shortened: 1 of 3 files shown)


piker/ui/_ohlc.py
```diff
@@ -421,3 +399,28 @@ def boundingRect(self):
             )
 
         )
+
+    def paint(
```
goodboy, line 403: This was just a method move to put `.paint()` at the end of the class def like the OHLC curve type.

piker/ui/_ohlc.py
```diff
@@ -146,7 +146,7 @@ def path_arrays_from_ohlc(
         # specifies that the first edge is never connected to the
         # prior bars last edge thus providing a small "gap"/"space"
         # between bars determined by ``bar_gap``.
-        c[istart:istop] = (0, 1, 1, 1, 1, 1)
+        c[istart:istop] = (1, 1, 1, 1, 1, 0)
```
goodboy, line 149: Hmm i feel like this got reverted since they fixed the bug in `pyqtgraph` that caused me to change this in the first place?

goodboy: Actually on latest `pg` master we seem to be good.
So I think all the issues I noticed are good now?

goodboy: Added a dev branch pin back to `requirements.txt`; prolly a better approach for now anyway..

piker/ui/_curve.py
```diff
@@ -56,17 +137,33 @@ def update_from_array(
...
+                finiteCheck=False,
             )
             profiler('generate fresh path')
 
+            # if self._step_mode:
+            #     self.path.closeSubpath()
+
         # TODO: get this working - right now it's giving heck on vwap...
         # if prepend_length:
         #     breakpoint()
```
goodboy, line 169: drop this lul.

piker/ui/_curve.py
```diff
@@ -83,21 +180,47 @@ def update_from_array(
...
-            # self.path.moveTo(new_x[0], new_y[0])
-            # self.path.connectPath(append_path)
-            self.path.connectPath(append_path)
+
+            path = self.path
+
+            # other merging ideas:
+            # https://stackoverflow.com/questions/8936225/how-to-merge-qpainterpaths
+            if self._step_mode:
+                if self._fill:
```
goodboy, line 210: We need to try out work from: https://github.com/pyqtgraph/pyqtgraph/pull/2032

piker/ui/_curve.py
```diff
@@ -112,8 +235,23 @@ def update_from_array(
         self.xData = x
         self.yData = y
 
-        self._xrange = x[0], x[-1]
-        self._last_line = QtCore.QLineF(x[-2], y[-2], x[-1], y[-1])
+        x0, x_last = self._xrange = x[0], x[-1]
+        y_last = y[-1]
+
+        if self._step_mode:
```
goodboy, line 241: prolly needs a comment to explain?

piker/ui/_curve.py
goodboy: Ok, tried it and it's still horrid. Gonna comment out this code and leave comments but honestly i think the consensus from the ranks is that that unfilled (except current bar) looks better anyway.

`pg` is using the same pain call:
https://github.com/pyqtgraph/pyqtgraph/pull/2032/files#diff-0110e152aa13c81b6b8b9c227f40b855c44cc86cf1fb5c9861d57e630898d48fR764 which is the super slow part afaict.

We can always take a look at this down the road if there's a big want from future users.

Issue #95, opened 2020-06-04
Pull crypto data from kraken


goodboy:
They have a public api that requires no key or account:
- public real-time quotes from [websockets, docs](https://docs.kraken.com/websockets/)
- [public endpoints](https://www.kraken.com/features/api#public-market-data) (require no api key) such as the [asset pair lookup api](https://www.kraken.com/features/api#get-tradable-pairs)

goodboy:
I've been able to get historical OHLC and and an "all symbols" query going quickly.
PR for client will come shortly.

PR #247, opened 2022-01-19
`marketstore` OHCLV storage and client integration

### LINKED ISSUE #143: Supervise the store

guilledk:
Update outdated `ingest` backend and replace the in-repo async wrapped marketstore client with the new:
https://github.com/pikers/anyio-marketstore

Includes support for a `tractor` based supervisor actor and thus resolves #143 which was merged in via the sub-PR #284.

---
#### ingest TODOs:
- [x] re-add `destroy` cli piker cmd (now implemented as `Storage.delete_ts()`)
- [ ] tick ingest to `marketstore` from `brokerd` feeds:
  - [ ] experiment with `techtonicdb` schema and if it can be used with the [aggregator plugin](https://github.com/alpacahq/marketstore/tree/master/contrib/ondiskagg) (got a feeling we'll need to write at least a `1Sec` bucket in order for this to work looking at the code:
    - [`aggregate()` calls into `model.FromTrades()`](https://github.com/alpacahq/marketstore/pull/399/files#diff-392952832ca30145443aff46786de412b7b5876a33e5dbcbb34dde20637c6370R285) and [`FromTrades()` only accepts a min step of `1Sec`](https://github.com/alpacahq/marketstore/pull/399/files#diff-cf7320bc8ecabe28fa6f7712d8df43143d3fa8577667f777c792db773ea5fbdbR192-R205)
    - we might be able to either write a golang plugin/extension to do it from pure tick data or we can make this part of a piker actor?
- [x] ~ohlcv ingest and load from multiple backends:~ this was added as part of the new history in #308.
  - `ib` seems like the natural place to start since `1s` is already offered in their history
    - we'll need separate series for dark vs. regular vlm
    - we'll likely want to write a diffing and syncing system to validate new captured histories from real-time runs (which almost always are slightly different)
  - `kraken` has tick history which has been a [todo for a while](#95)
- [ ] benchmark binance tick feed against [binancefeeder](https://github.com/alpacahq/marketstore/tree/master/contrib/binancefeeder)

---
#### `_ahab` TODO:
 - [ ] probably deliver back net-socket info over the `ctx.started()` call?
   - [ ] grprc socket
   - [ ] ws addr
 - [x] cli support for running `pikerd` with the tsdb stuff spawned?
   - [x] maybe a `--tsdb` or `--data` or something?
 - [ ] figure out what to do with the `mkts.yaml` config?
  - step by step interaction test lisit:
   - [x] `pikerd --tsdb` should raise `DockerNotStarted` and appropriate perms error on no `sudo` (for now)
   - [x] ctrl-c should kill container instance
     - [ ] this **does not really work** when run with `sudo` since we relinquish root perms in order to create shms..

### DIFF
--- a/piker/cli/__init__.py
+++ b/piker/cli/__init__.py
@@ -54,13 +54,19 @@ def pikerd(loglevel, host, tl, pdb, tsdb):
             trio.open_nursery() as n,
         ):
             if tsdb:
+                # TODO:
+                # async with maybe_open_marketstored():
+
                 from piker.data._ahab import start_ahab
                 log.info('Spawning `marketstore` supervisor')
-                ctn_ready = await n.start(start_ahab)
+                ctn_ready = await n.start(
+                    start_ahab,
+                    'marketstored',
+                )
                 await ctn_ready.wait()
                 log.info('`marketstore` container:{uid} up')
 
-            await trio.sleep_forever()
+                await trio.sleep_forever()
 
     trio.run(main)
 
--- a/piker/cli/__init__.py
+++ b/piker/cli/__init__.py
@@ -50,7 +50,7 @@ def pikerd(loglevel, host, tl, pdb, tsdb):
             open_pikerd(
                 loglevel=loglevel,
                 debug_mode=pdb,
-            ) as services,
+            ),  # normally delivers a ``Services`` handle
             trio.open_nursery() as n,
         ):
             if tsdb:
--- a/piker/cli/__init__.py
+++ b/piker/cli/__init__.py
@@ -66,7 +66,7 @@ def pikerd(loglevel, host, tl, pdb, tsdb):
                 await ctn_ready.wait()
                 log.info('`marketstore` container:{uid} up')
 
-                await trio.sleep_forever()
+            await trio.sleep_forever()
 
     trio.run(main)
 
--- a/piker/cli/__init__.py
+++ b/piker/cli/__init__.py
@@ -22,9 +22,16 @@ DEFAULT_BROKER = 'questrade'
 @click.option('--tl', is_flag=True, help='Enable tractor logging')
 @click.option('--pdb', is_flag=True, help='Enable tractor debug mode')
 @click.option('--host', '-h', default='127.0.0.1', help='Host address to bind')
-def pikerd(loglevel, host, tl, pdb):
-    """Spawn the piker broker-daemon.
-    """
+@click.option(
+    '--tsdb',
+    is_flag=True,
+    help='Enable local ``marketstore`` instance'
+)
+def pikerd(loglevel, host, tl, pdb, tsdb):
+    '''
+    Spawn the piker broker-daemon.
+
+    '''
     from .._daemon import open_pikerd
     log = get_console_log(loglevel)
 
@@ -38,7 +45,21 @@ def pikerd(loglevel, host, tl, pdb):
         ))
 
     async def main():
-        async with open_pikerd(loglevel=loglevel, debug_mode=pdb):
+
+        async with (
+            open_pikerd(
+                loglevel=loglevel,
+                debug_mode=pdb,
+            ) as services,
+            trio.open_nursery() as n,
+        ):
+            if tsdb:
+                from piker.data._ahab import start_ahab
+                log.info('Spawning `marketstore` supervisor')
+                ctn_ready = await n.start(start_ahab)
+                await ctn_ready.wait()
+                log.info('`marketstore` container:{uid} up')
+
             await trio.sleep_forever()
 
     trio.run(main)
--- a/piker/cli/__init__.py
+++ b/piker/cli/__init__.py
@@ -16,20 +16,6 @@ from .. import config
 log = get_logger('cli')
 DEFAULT_BROKER = 'questrade'
 
-_config_dir = click.get_app_dir('piker')
-_watchlists_data_path = os.path.join(_config_dir, 'watchlists.json')
-_context_defaults = dict(
-    default_map={
-        # Questrade specific quote poll rates
-        'monitor': {
-            'rate': 3,
-        },
-        'optschain': {
-            'rate': 1,
-        },
-    }
-)
-
 
 @click.command()
 @click.option('--loglevel', '-l', default='warning', help='Logging level')
@@ -58,7 +44,7 @@ def pikerd(loglevel, host, tl, pdb):
     trio.run(main)
 
 
-@click.group(context_settings=_context_defaults)
+@click.group(context_settings=config._context_defaults)
 @click.option(
     '--brokers', '-b',
     default=[DEFAULT_BROKER],
@@ -87,8 +73,8 @@ def cli(ctx, brokers, loglevel, tl, configdir):
         'loglevel': loglevel,
         'tractorloglevel': None,
         'log': get_console_log(loglevel),
-        'confdir': _config_dir,
-        'wl_path': _watchlists_data_path,
+        'confdir': config._config_dir,
+        'wl_path': config._watchlists_data_path,
     })
 
     # allow enabling same loglevel in ``tractor`` machinery
--- a/piker/config.py
+++ b/piker/config.py
@@ -17,6 +17,8 @@
 """
 Broker configuration mgmt.
 """
+import platform
+import sys
 import os
 from os.path import dirname
 import shutil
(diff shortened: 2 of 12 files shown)


goodboy:
@guilledk can you rebase this amigo?

goodboy:
Merged in the super stuff from #284 and will begin 1s OHLC ingest from ib shortly 🏄🏼

goodboy:
Oh right and I have to go through the todos and update!

Mostly just want get a first set of eyes on the code.

piker/data/_ahab.py
```diff
@@ -133,6 +133,136 @@ def unpack_msg(err: Exception) -> str:
                 c.kill()
 
 
+class Container:
```
guilledk, line 136: What you think of creating a `docker-utils` python package with a lot of common abstractions to deal with dockerisms, it would be diferent than `pytest-dockerctl` cause it wouldnt be only about pytest, I have a bunch of utils id like in a pip instalable package somewhere

guilledk: Examples: https://github.com/guilledk/py-eosio/blob/71065aad8a792a54b05f63b8ce8b4ac68b00b1e0/py_eosio/sugar.py#L338

`wait_for_attr`, `get_container`, `docker_open_process`, `docker_wait_process`, `docker_move_into`

goodboy: i think this is an amazing idea!

for example here, all the supervisor stuff i had to hack would be great to get into a (somewhat) formalized api layer.

goodboy: Yeah I'm also not settled on a lot of this api / naming so don't take it too too seriously and feel free to suggest changes.

I'm also fine with landing this code and making it possible for someone else to go through and do a cleanup / reorg if desired.

I mostly just want to get in the ability to store large data sets for users and prime things for proper real-time ingest from tick feeds.

piker/cli/__init__.py
```diff
@@ -16,29 +16,22 @@
...
 @click.command()
 @click.option('--loglevel', '-l', default='warning', help='Logging level')
 @click.option('--tl', is_flag=True, help='Enable tractor logging')
 @click.option('--pdb', is_flag=True, help='Enable tractor debug mode')
 @click.option('--host', '-h', default='127.0.0.1', help='Host address to bind')
-def pikerd(loglevel, host, tl, pdb):
-    """Spawn the piker broker-daemon.
-    """
+@click.option(
+    '--tsdb',
```
goodboy, line 26: So this is what a user passes to indicate `pikerd` will spawn the `marketstore` container in a subactor.

piker/config.py
```diff
@@ -17,21 +17,109 @@
...
 
 from .log import get_logger
 
 log = get_logger('broker-config')
 
-_config_dir = click.get_app_dir('piker')
+
+# taken from ``click`` since apparently they have some
+# super weirdness with sigint and sudo..no clue
+def get_app_dir(app_name, roaming=True, force_posix=False):
```
goodboy, line 37: Ahh right, yeah this is probably confusing af..

It seems that using SIGIN cancellation was somehow super oddly being swallowed/ignored by `click` and so pulling out this one routine we needed (which probably isn't a horrible idea to maintain on our own anyway?) was the fast hack.

piker/data/_ahab.py
```diff
@@ -0,0 +1,469 @@
...
+        # async with ctx.open_stream() as stream:
+
+        try:
+
+            # TODO: we might eventually want a proxy-style msg-prot here
+            # to allow remote control of containers without needing
+            # callers to have root perms?
+            await trio.sleep_forever()
+
+            # await cntr.cancel()
```
goodboy, line 347: Left over code from when I thought the issues with "graceful cancellation" were to do with root perms stuff. Turns out it wasn't that and it was the actual `marketstore` app layer not shutting down. 

I will pull out this cruft shortly.

piker/data/_ahab.py
```diff
@@ -0,0 +1,469 @@
...
+    task_status: TaskStatus[trio.Event] = trio.TASK_STATUS_IGNORED,
+
+) -> None:
+    '''
+    Start a ``docker`` container supervisor with given service name.
+
+    Currently the actor calling this task should normally be started
+    with root permissions (until we decide to use something that doesn't
+    require this, like docker's rootless mode or some wrapper project) but
+    te root perms are de-escalated after the docker supervisor sub-actor
```
goodboy, line 390: s/te/the

piker/data/_source.py
```diff
@@ -254,61 +253,6 @@ def iterfqsns(self) -> list[str]:
         return keys
 
 
-def from_df(
```
goodboy, line 257: ahh yeah right!

we can drop this (and maybe `pandas` altogether for now) since i rewrote the `ib` ohlc frame parser to just cast directly to `numpy`.

piker/data/cli.py
```diff
@@ -49,51 +57,58 @@
...
-                for sym in names:
-                    # tbk = _tick_tbk.format(sym)
-                    tbk = tuple(sym, *_tick_tbk_ids)
-                    print(f"Destroying {tbk}..")
-                    await client.destroy(mk_tbk(tbk))
-            else:
-                print("Nothing deleted.")
 
-    tractor.run(main)
+# @cli.command()
```
goodboy, line 78: ahh yeah we can probably drop this.

the way i'd like to deal with "tsdb management" is the new `storesh` repl (with embedded `ipython`) which will have a small interactive API for manual db tinkering.

piker/data/feed.py
```diff
@@ -188,6 +191,22 @@ async def _setup_persistent_brokerd(
         await trio.sleep_forever()
 
 
+async def start_backfill(
```
goodboy, line 194: Ok so there is a bunch of new code for this in #302 that hasn't yet been factored out. I wasn't sure if that should all come as part of a "backend ohlc history support" PR (aka more of less what #305 is) or be included here?

Technically this PR will land with minimal backend support (i think just `ib`?) so it's more about how much code ya'll want test review :joy:

goodboy:
Replaced by #308 🏄🏼

piker/data/feed.py
goodboy: Yeah this kinda stuff was the main reason for dropping this PR and just aggregating it all in to #308.

piker/data/_ahab.py
goodboy: Done in #308.

PR #249, opened 2022-01-23
Basic vlm display

goodboy:
This pulls out the base functionality to get a volume subchart from #231.

The purpose of splitting out this work is,
- to make code review simpler and demonstrate the initial UX changes and graphics implementation in use
- create **clear a demarcation commit** to compare and benchmark versus the new FSP driven approach eventually built out in #231 
- get users testing before bringing in more complex `pyqtgraph` extensions which were developed as part of #231 and pertain to what will be shared aggregated feeds on a single chart

Please feel free to check out this branch and run it through the wringer 😂

### DIFF
--- a/piker/ui/_chart.py
+++ b/piker/ui/_chart.py
@@ -25,6 +25,9 @@ from PyQt5.QtCore import Qt
 from PyQt5.QtWidgets import (
     QFrame,
     QWidget,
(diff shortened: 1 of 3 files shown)


guilledk (approved):
Runs great on my machine!

PR #253, opened 2022-01-24
Dolla vlm

goodboy:
First draft of so called "dollar volume" FSP added as a 3rd subchart for comparison with unit volume.

You can find more info on this measure here:
- example implementation using `ohl3` [on tinaview](https://www.tradingview.com/script/O3xqRCCi-Total-Dollar-Volume/)
- related DVL def [from investopedia](https://www.investopedia.com/terms/d/dollar-volume-liquidity.asp)

We currently use the `ohlc3` "typical price" mean to calculate the price part for historical bars but use the real-time `price * size` for each incoming tick.

This is again history factored from #231 and is meant as a small reviewable patchset.
When the final history from that PR is landed $vlm and vlm will be overlayed in a single subchart 😎

### DIFF
--- a/piker/fsp/_engine.py
+++ b/piker/fsp/_engine.py
@@ -34,7 +34,7 @@ from ..data import attach_shm_array
 from ..data.feed import Feed
 from ..data._sharedmem import ShmArray
 from ._momo import _rsi, _wma
-from ._volume import _tina_vwap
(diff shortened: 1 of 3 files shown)


goodboy:
Oh, also note the impl doesn't include leverage adjustments for the instrument yet (i.e. futures aren't properly scaled).

guilledk (approved):

PR #259, opened 2022-01-24
Overlayed $vlm

goodboy:
Final touches on getting *overlayed unit and dollar* volume curves as our default sub-chart 🏄🏼 

The main bit of fancyness comes from the new `humanize()`d axes values and a single curve which starts as the unit vlm until the $vlm fsp is fully spawned at which point the curve is flipped to the latter. Also includes nice minimal labels for each y-axis below in the space available to the x-axis 😎

### DIFF
--- a/piker/fsp/_momo.py
+++ b/piker/fsp/_momo.py
@@ -1,5 +1,5 @@
 # piker: trading gear for hackers
-# Copyright (C) 2018-present  Tyler Goodlet (in stewardship of piker0)
+# Copyright (C) Tyler Goodlet (in stewardship of pikers)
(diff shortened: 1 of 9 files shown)


guilledk (approved):

goodboy:
Make sure y'all test this one thoroughly because a lot was factored out of the original dev branch (#231) and I was to be sure the end product is the same from that one as it is here 😂

goodboy:
Updated our `pyqtgraph` fork pin as per #262 so I think this is good to land.

@iamzoltan you gonna test er wut?

PR #260, opened 2022-01-25
Dark vlm

goodboy:
WIP plotting "dark" volume trades overlayed with normal feed provided by the `ib` backend.

The feed for these are the so called *unreportable trades* as vaguely mentioned in their tick types under id `48`:
https://interactivebrokers.github.io/tws-api/tick_types.html

> Last trade details (Including both "Last" and "Unreportable Last" trades).

---
ToDo:
- [x] separate curve for dark volume overlaid with our current $vlm
- ~[ ] possibly a line overlayed with the OHLC bars to some how indicate the *direction* of these ticks~ (deferred to #266)

### DIFF
--- a/piker/fsp/_momo.py
+++ b/piker/fsp/_momo.py
@@ -23,6 +23,7 @@ from typing import AsyncIterator, Optional
 import numpy as np
 from numba import jit, float64, optional, int64
 
+from ._api import fsp
 from ..data._normalize import iterticks
 from ..data._sharedmem import ShmArray
 
@@ -106,7 +107,7 @@ def ema(
 #     nopython=True,
 #     nogil=True
 # )
-def rsi(
+def _rsi(
 
     # TODO: use https://github.com/ramonhagenaars/nptyping
     signal: 'np.ndarray[float64]',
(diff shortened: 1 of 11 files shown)


piker/fsp/_momo.py
```diff
@@ -169,10 +170,8 @@ def wma(
     return np.convolve(signal, weights, 'valid')
 
 
-# @piker.fsp.emit(
-#     timeframes=['1s', '5s', '15s', '1m', '5m', '1H'],
-# )
-async def _rsi(
+@fsp
```
guilledk, line 173: Slick API!

guilledk (approved):

goodboy:
One last thing is we gotta land https://github.com/goodboy/tractor/pull/295 for the new `NamespacePath` type.

Ping @guilledk

piker/fsp/_momo.py
goodboy: 💋 thanks amigo

goodboy:
One last note on this was whether to *always* show the dark vlm curve even if no feed is available from the data provider. For now it's just getting set to zero and that zero curve is updated on each step; we should probably have some kind of "symbol info" field that determines whether we draw it at all to avoid the unneeded overhead in most cases.

PR #268, opened 2022-02-02
Trade ratez

goodboy:
WIP displaying overlayed "flow rates" on top of our volume fsp subchart 😎 

Basically the idea here is to plot measure flow in windowed metric terms to give a view into HFT activity as well as compare rates across instruments/markets.

Currently we're displaying the values provided verbatim by IB [ticks `294` & `295`](https://interactivebrokers.github.io/tws-api/tick_types.html) which afaik are also available to be used in their [advanced scanner](https://www1.interactivebrokers.com/en/software/tws/usersguidebook/technicalanalytics/market_scanner_types.htm) and its [API accessible params](https://interactivebrokers.github.io/tws-api/scanner_parameters.html).

We're currently accessing these ticks via the pre-packing which `ib_insync` does in their [`Ticker` type](https://ib-insync.readthedocs.io/api.html#ib_insync.ticker.Ticker.tradeRate).

For reference on what each metrics is onstensibly measuring:
- *Top Trade Rate*:  Contracts with the highest number of trades in the past 60 seconds (regardless of the sizes of those trades). Displays the trades/minute field.
- *Top Volume Rate*:  The top volume rate per minute
- both of these `ib` provided values appear to only update every 25s from my eye-balling of the curves.

---
TODO:
- [x] our own real-time, by bars period version of both of these rates
  - ~[ ] a `period:` control on the side pane? or can we wait for a next PR for this?~ -> #279
- [x] a $vlm version (though not sure that shows anything more useful?
- [ ] figure out how much more useful *trade* rates vs. vlm is at detecting HFT activity
- ~[ ] a legend which makes it more clear which curve is which since the vlm subchart is getting a little full now 😂~ -> #279

### DIFF
--- a/piker/fsp/_volume.py
+++ b/piker/fsp/_volume.py
@@ -22,8 +22,15 @@ from tractor.trionics._broadcast import AsyncReceiver
 from ._api import fsp
 from ..data._normalize import iterticks
 from ..data._sharedmem import ShmArray
+from ._momo import _wma
+from ..log import get_logger
+
+log = get_logger(__name__)
 
 
+# NOTE: is the same as our `wma` fsp, and if so which one is faster?
+# Ohhh, this is an IIR style i think? So it has an anchor point
+# effectively instead of a moving window/FIR style?
 def wap(
 
     signal: np.ndarray,
@@ -131,8 +138,6 @@ async def dolla_vlm(
     chl3 = (a['close'] + a['high'] + a['low']) / 3
     v = a['volume']
 
-    from ._momo import wma
-
     # on first iteration yield history
     yield {
         'dolla_vlm': chl3 * v,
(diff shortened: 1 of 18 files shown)


piker/fsp/_volume.py
```diff
@@ -174,4 +202,132 @@ async def dolla_vlm(
             # tina_lvlm = c+h+l/3 * v
             # print(f' tinal vlm: {tina_lvlm}')
 
-            yield key, output
+
+@fsp(
```
goodboy, line 206: This is the primary new addition to the built-in fsp set 😎

piker/fsp/_volume.py
```diff
@@ -174,4 +202,132 @@ async def dolla_vlm(
...
+        yield 'dark_dvlm_rate', dark_dvlm_wma[-1]
+
+        dark_trade_rate_wma = _wma(
+            dvlm_shm.array['dark_trade_count'],
+            period,
+            weights=weights,
+        )
+        yield 'dark_trade_rate', dark_trade_rate_wma[-1]
+
+        # XXX: ib specific schema we should
```
goodboy, line 320: This is commented out rates from ib that i'm not sure we'll even want since as mentioned they're only updated every 25s..

guilledk (approved):

PR #271, opened 2022-02-06
Ib mkt closed

goodboy:
WIP fixes to make it possible to load OHLC history even when markets are closed.

We haven't yet done any introspection of symbol "market open hours" and instead have relied on whether or not a real-time quote arrives before starting up all data feed machinery. This patch instead allows loading history after a timeout so that the UI doesn't just hang 😂 

ping @wattygetlood

### DIFF
--- a/piker/brokers/ib.py
+++ b/piker/brokers/ib.py
@@ -507,6 +507,7 @@ class Client:
     async def get_quote(
         self,
         symbol: str,
+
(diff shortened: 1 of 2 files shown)


goodboy:
I've made follow up issue #276 to get this fully working, but for now I think this change is minor enough to just land and it at least does load price history during off hours.

PR #278, opened 2022-02-08
Windows fixes yo

wattygetlood:


guilledk (approved):
Good work!

piker/brokers/ib.py
```diff
@@ -1204,6 +1204,11 @@ async def backfill_bars(
     https://github.com/pikers/piker/issues/128
 
     """
+    if platform.system() == 'Windows':
+        log.warning(
+            'Decreasing history query count to 4 since, windows...')
+        count = 4
```
goodboy, line 1210: It will be nice to avoid this once we have replicated history via tsdb.

piker/ui/_window.py
```diff
@@ -153,8 +153,8 @@ class MainWindow(QtGui.QMainWindow):
     # XXX: for tiling wms this should scale
     # with the alloted window size.
     # TODO: detect for tiling and if untrue set some size?
-    # size = (300, 500)
-    size = (0, 0)
+    size = (300, 500)
+    #size = (0, 0)
```
goodboy, line 157: drop this

goodboy:
Ok i fixed a buncha stuff that wasn't so nice on linux + made the search results "view" sizing dynamic to the largest result.

My last question for y'all would be should we **always dynamically size** the results view to the contents or just leave it like i have it where it just grows to the largest one see thus far?

piker/ui/_search.py
```diff
@@ -153,23 +158,55 @@ def set_font_size(self, size: int = 18):
...
+            self.setMaximumSize(self.width(), rows * row_px)
+            self.setMinimumSize(self.width(), rows * row_px)
+
+        # self.resize(self.width(), rows * row_px)
+        self.resize(self.width(), rows * row_px)
+        self._last_window_h = window_h
+
+        # size to width of longest result seen thus far
+        # TODO: should we always dynamically scale to longest result?
+        if self.width() < col_w_tot:
```
goodboy, line 206: As per my comment, we should just always size the largest result it just makes the UX a little *jittery* when results are filling out 😂 

I'm not sure which i like better.
I guess we could also add a bit of state tracking between each `.show()` of the search such that it only grows from the last "session" (aka since showing the widget) of results?

I dunno. give feeback!

PR #284, opened 2022-02-16
Ahab's dick

### LINKED ISSUE #143: Supervise the store

goodboy:
Add's a simple `docker` supervisor for spawning containers and launching `marketstore` in the actor tree using a `@tractor.context` endpoint.

Resolves #143.

In its current state you can start the container with `python -m piker.data._ahab` from the root repo dir.

---
#### `_ahab` TODO:  (**UPDATE: MOVED TO #247**)
 - [ ] probably deliver back net-socket info over the `ctx.started()` call?
   - [ ] grprc socket
   - [ ] ws addr
 - [ ] cli support for running `pikerd` with the tsdb stuff spawned?
   - [ ] maybe a `--tsdb` or `--data` or something?
 - [ ] figure out what to do with the `mkts.yaml` config?

### DIFF
--- a/piker/config.py
+++ b/piker/config.py
@@ -17,6 +17,8 @@
 """
 Broker configuration mgmt.
 """
+import platform
+import sys
 import os
 from os.path import dirname
 import shutil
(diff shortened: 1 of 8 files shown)


piker/config.py
```diff
@@ -17,21 +17,109 @@
...
 
 from .log import get_logger
 
 log = get_logger('broker-config')
 
-_config_dir = click.get_app_dir('piker')
+
+# taken from ``click`` since apparently they have some
+# super weirdness with sigint and sudo..no clue
+def get_app_dir(app_name, roaming=True, force_posix=False):
```
goodboy, line 37: Literally had to add this because use of `click` somehow causes `SIGINT` to be masked.

See https://github.com/goodboy/tractor/issues/302

piker/data/_ahab.py
```diff
@@ -0,0 +1,222 @@
...
+import json
+from docker.models.containers import Container
+
+from ..log import get_logger  # , get_console_log
+from ..config import _config_dir
+
+log = get_logger(__name__)
+
+
+_config = '''
```
goodboy, line 40: This is the config we need to talk about.

guilledk (approved):

PR #286, opened 2022-02-23
update windows install instructions

sirdinkus:
just updating the guide for windows users to streamline the process as well as setting up vscode


goodboy:
Live readme render can be seen [here](https://github.com/pikers/piker/tree/wattygetlood-patch-1#install-for-tinas) which isn't looking great 😂 

@wattygetlood we need use of some code blocks and better bullet formatting i'd say.

You can also try out your changes on an rst renderer like this: https://livesphinx.herokuapp.com/

README.rst
```diff
@@ -74,23 +74,57 @@ for a development install::
 
 install for tinas
 *****************
-for windows peeps you can start by getting `conda installed`_
-and the `C++ build toolz`_ on your system.
+for windows peeps you can start by installing all the prerequisite software:
 
-then, `crack a conda shell`_ and run the following commands::
+- install git with all default settings - https://git-scm.com/download/win
```
goodboy, line 79: so i'd recommend doing a hyperlink syntax here instead of the explicit urls, though if you think that's a little 🤯 for tinas then this is fine too.

basically you'd do something like

```markdown
- [install git with all default settings](https://git-scm.com/download/win)
- [install anaconda all default settings](https://www.anaconda.com/products/individual)
...
etc.
etc.
```

README.rst
```diff
@@ -74,23 +74,57 @@ for a development install::
...
+    conda create -n pikonda # creates conda environment named pikonda
+    conda activate pikonda # activates pikonda
+    
+    conda install -c conda-forge python-levenshtein # in case it is not already installed
+    conda install pip # may already be installed
+    pip # will show if pip is installed
+    
+    pip install -e . -r requirements.txt # install piker in editable mode
+
+test Piker to see if it is working::
```
goodboy, line 101: change this to be `piker`

goodboy (commented):
@sirdinkus looking decent, i have a couple recos to make it look a little more "docs-y" as mentioned in the diff comments.

also remember you can look at the md rendered version on your GH [branch page](https://github.com/pikers/piker/tree/wattygetlood-patch-1#install-for-tinas) 😉

README.rst
sirdinkus: I personally like seeing the url so I know what I am clicking on.

sirdinkus:
@goodboy I think it's probably good enough to merge now, so there is a functional installation instructions for windows.

guilledk (approved):

Issue #93, opened 2020-05-26
marketstore integration


goodboy:
It works, that [marketstore](https://github.com/alpacahq/marketstore), and I gotta say it's not too bad.

I've got tick data from real-time feeds finally going and it seems pretty aight for latency even with the `numpy.array()` write inline with `trio`/`tractor` streaming code.
 
Need to work on the schema a bit and figure out if using [variable length](https://github.com/alpacahq/marketstore/blob/master/executor/writer.go#L79) column series is indeed a better fit.

Here's the outstanding todos:
-  use [trio-websocket](https://github.com/hyperiongray/trio-websocket) to stream from the [streaming trigger](https://github.com/alpacahq/marketstore/tree/master/contrib/stream) system
- automate the official [docker image](https://github.com/alpacahq/marketstore#docker) spinup using either [docker-py](https://docker-py.readthedocs.io/en/stable/) or [aiodocker](https://github.com/aio-libs/aiodocker)
- capture and process the [structured logging](https://github.com/NLog/NLog/wiki/How-to-use-structured-logging) from [the server](https://github.com/alpacahq/marketstore/blob/master/utils/log/log.go#L6) which is using [zap](https://pkg.go.dev/go.uber.org/zap@v1.15.0?tab=doc) using a monitoring actor I guess?

PR #287, opened 2022-02-28
Async hist loading

goodboy:
Re-work of the `piker.data.feed` machinery to do async loading of both real-time and historical data for faster chart startup and a better setup to get the work from #93 landed.

Part of this work discovered the unexpected behavior that is https://github.com/python-trio/trio/issues/2258.

Included are a bunch of other changes summarized by "topic":

---
#### miscellaneous fixups
- docs strings added to curve graphics code
- changed display loop throttle warnings to only fire on >= display rate cycles
- dropped stale deps on `msgpack` and `msgpack-numpy`
- place the sample period (aka "time frame") in the window header
- add an `open_piker_runtime()` to make it easy to write a client that uses all the `tractor` apis correctly

---
#### data feed rework
- drop passing `ShmArray`s to the backend `stream_quotes()` endpoints since they don't manually write themselves
- allow using `open_feed()` without requiring a real-time quote stream (eg. just for accessing shm / history)
- drop the dedicated task to update graphics on sample period steps, instead just let the display loop detect new samples and update on the next event
- expect fsps to request the sample period for incrementing shm history based on calculating the step size from source data
- don't crash (just warn) on a double remove of a sampled fee subscription

---
#### misc graphics
- make `ChartPlotWidget.increment_view()` take a `steps: int`
- don't zero clearing rate fsp curves on a new sample step
- fix the display loop's `mamin()` range calcer to not crash on no-data-in-view

### DIFF
--- a/piker/_daemon.py
+++ b/piker/_daemon.py
@@ -80,7 +80,6 @@ class Services(BaseModel):
         ) -> Any:
 
             with trio.CancelScope() as cs:
-
                 async with portal.open_context(
                     target,
                     **kwargs,
@@ -89,19 +88,17 @@ class Services(BaseModel):
 
                     # unblock once the remote context has started
                     task_status.started((cs, first))
-
+                    log.info(
+                        f'`pikerd` service {name} started with value {first}'
+                    )
                     # wait on any context's return value
                     ctx_res = await ctx.result()
-                    log.info(
-                        f'`pikerd` service {name} started with value {ctx_res}'
-                    )
 
                 # wait on any error from the sub-actor
                 # NOTE: this will block indefinitely until cancelled
-                # either by error from the target context function or
-                # by being cancelled here by the surroundingn cancel
-                # scope
-                return await (portal.result(), ctx_res)
+                # either by error from the target context function or by
+                # being cancelled here by the surrounding cancel scope
+                return (await portal.result(), ctx_res)
 
         cs, first = await self.service_n.start(open_context_in_task)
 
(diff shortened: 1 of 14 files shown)


piker/_daemon.py
```diff
@@ -447,3 +494,25 @@ async def maybe_open_emsd(
 
     ) as portal:
         yield portal
+
+
+# TODO: ideally we can start the tsdb "on demand" but it's
```
goodboy, line 499: Obvs coming down the road as part of what results from #247, left in as a guide.

piker/fsp/_engine.py
```diff
@@ -381,14 +380,19 @@ async def poll_and_sync_to_step(
 
             s, step, ld = is_synced(src, dst)
 
+            # detect sample period step for subscription to increment
+            # signal
+            times = src.array['time']
```
goodboy, line 385: We'll probably want a standard api for this eventually?

piker/data/feed.py
```diff
@@ -86,14 +106,28 @@ class _FeedsBus(BaseModel):
...
+        *args,
+
+    ) -> None:
+
+        async def start_with_cs(
+            task_status: TaskStatus[
+                trio.CancelScope] = trio.TASK_STATUS_IGNORED,
+        ) -> None:
+            with trio.CancelScope() as cs:
+                self.nursery.start_soon(
```
goodboy, line 121: I think we should probably move this to be a `self.nusery.start()` for now until https://github.com/python-trio/trio/issues/2258 is resolved.

piker/data/feed.py
```diff
@@ -128,30 +162,108 @@ def get_feed_bus(
...
+    '''
+    Load and manage historical data including the loading of any
+    available series from `marketstore` as well as conducting real-time
+    update of both that existing db and the allocated shared memory
+    buffer.
+
+    '''
+    # TODO: history retreival, see if we can pull from an existing
+    # ``marketstored`` daemon
+    log.info('Scanning for existing `marketstored`')
```
goodboy, line 213: Woops, this should probably get commented.
Obviously these imports will come in with #247.

piker/data/feed.py
```diff
@@ -128,30 +162,108 @@ def get_feed_bus(
...
+
+    opened = we_opened_shm
+    # TODO: history validation
+    # assert opened, f'Persistent shm for {symbol} was already open?!'
+    # if not opened:
+    #     raise RuntimeError("Persistent shm for sym was already open?!")
+
+    if opened:
+        if arrays:
+            await tractor.breakpoint()
```
goodboy, line 229: just placeholder for #247 work.

guilledk (approved):

PR #295, opened 2022-04-08
Fqsns

goodboy:
`fqsn` = "fully qualified symbol name"

Basically the idea here is for all client-ish code (aka programs/actors that ask the provider agnostic layers in the stack for data) should be able to tell which backend / venue / derivative each data feed/flow is from by an explicit string key of the current form `<instrumentname>.<venue>.<suffixwithmetadata>.<brokerbackendname>`.

I have thoughts that we should actually change this to be more like an "attr lookup" (like how the web should have done urls, but marketting peeps ruined it etc. etc.):

`<broker>.<venue>.<instrumentname>.<suffixwithmetadata>`

This preps for new derivatives search support in the `ib` backend coming in #294 (which will get rebased and likely renamed after this).

----
Moar,

- there's some fixes to the data (feed) layer to be more tolerant to IPC failures
- there might be some dependency custom fork changes i have to include, gotta dig through what's missing - CI and all..

### DIFF
--- a/piker/brokers/kraken.py
+++ b/piker/brokers/kraken.py
@@ -18,9 +18,9 @@
 Kraken backend.
 
 '''
-from contextlib import asynccontextmanager
+from contextlib import asynccontextmanager as acm
 from dataclasses import asdict, field
-from typing import Dict, List, Tuple, Any, Optional, AsyncIterator
+from typing import Any, Optional, AsyncIterator, Callable
 import time
 
 from trio_typing import TaskStatus
@@ -80,7 +80,7 @@ ohlc_dtype = np.dtype(_ohlc_dtype)
 _show_wap_in_history = True
 
 
-_symbol_info_translation: Dict[str, str] = {
+_symbol_info_translation: dict[str, str] = {
     'tick_decimals': 'pair_decimals',
 }
 
@@ -102,16 +102,16 @@ class Pair(BaseModel):
     lot_multiplier: float
 
     # array of leverage amounts available when buying
-    leverage_buy: List[int]
+    leverage_buy: list[int]
     # array of leverage amounts available when selling
-    leverage_sell: List[int]
+    leverage_sell: list[int]
 
     # fee schedule array in [volume, percent fee] tuples
-    fees: List[Tuple[int, float]]
+    fees: list[tuple[int, float]]
 
     # maker fee schedule array in [volume, percent fee] tuples (if on
     # maker/taker)
-    fees_maker: List[Tuple[int, float]]
+    fees_maker: list[tuple[int, float]]
 
     fee_volume_currency: str  # volume discount currency
     margin_call: str  # margin call level
@@ -153,7 +153,7 @@ class OHLC:
     volume: float  # Accumulated volume **within interval**
     count: int  # Number of trades within interval
     # (sampled) generated tick data
-    ticks: List[Any] = field(default_factory=list)
+    ticks: list[Any] = field(default_factory=list)
 
 
 def get_config() -> dict[str, Any]:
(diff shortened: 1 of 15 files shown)


goodboy:
Bleh, CI is actually failing due to this change:

```
requirements.txt (line 4))
  Cloning git://github.com/goodboy/tractor.git (to revision master) to ./src/tractor
  Running command git clone --filter=blob:none --quiet git://github.com/goodboy/tractor.git /home/runner/work/piker/piker/src/tractor
  fatal: remote error:
    The unauthenticated git protocol on port 9418 is no longer supported.
  Please see https://github.blog/2021-09-01-improving-git-protocol-security-github/ for more information.
```

So i guess even more hassle to test things 😂

piker/brokers/kraken.py
```diff
@@ -18,9 +18,9 @@
 Kraken backend.
 
 """
-from contextlib import asynccontextmanager
+from contextlib import asynccontextmanager as acm
```
iamzoltan, line 21: this is definitely cleaner

piker/clearing/_ems.py
```diff
@@ -20,7 +20,7 @@
 """
 from contextlib import asynccontextmanager
 from dataclasses import dataclass, field
-from math import isnan
+# from math import isnan
```
iamzoltan, line 23: does this still need to be here?

piker/clearing/_paper_engine.py
```diff
@@ -32,6 +32,7 @@
 
 from .. import data
 from ..data._normalize import iterticks
+from ..data._source import uncons_fqsn
```
iamzoltan, line 35: uncons should likely be changed to unpack

piker/data/_source.py
```diff
@@ -91,31 +92,113 @@ def ohlc_zeros(length: int) -> np.ndarray:
     return np.zeros(length, dtype=base_ohlc_dtype)
 
 
+def uncons_fqsn(fqsn: str) -> tuple[str, str, str]:
```
iamzoltan, line 95: change to unpack_fqsn

piker/clearing/_client.py
```diff
@@ -194,21 +196,22 @@ async def open_ems(
...
 
         async with (
-
             # connect to emsd
             portal.open_context(
 
                 _emsd_main,
-                broker=broker,
-                symbol=symbol.key,
+                fqsn=fqsn,
```
iamzoltan, line 214: some docs somewhere to explain what `fqsn` stands for would be welcomed

iamzoltan (commented):
looks good to me. mainly some docs on the new nomenclature as well as a func name change for clarity.

I dig it.

piker/clearing/_ems.py
goodboy: i don't think so, i'll scan the old comments to make sure.

piker/clearing/_paper_engine.py
goodboy: right, good call.

piker/clearing/_client.py
goodboy: yah no kidding 😂 

the description in this PR would probably me more or less sufficient yah?

iamzoltan: definitely, but it would be nice to have it in the code somewhere

goodboy: added as doc string on `Symbol.front_fqsn()`.

piker/data/_source.py
```diff
@@ -131,37 +214,44 @@ def front_feed(self) -> tuple[str, str]:
...
+    def front_fqsn(self) -> str:
+        '''
+        fqsn = "fully qualified symbol name"
 
-@validate_arguments
-def mk_symbol(
+        Basically the idea here is for all client-ish code (aka programs/actors
+        that ask the provider agnostic layers in the stack for data) should be
+        able to tell which backend / venue / derivative each data feed/flow is
+        from by an explicit string key of the current form:
```
iamzoltan, line 231: hot hot hot

iamzoltan (approved):
looks gouda to me my dude

guilledk (approved):

Issue #290, opened 2022-03-19
Kraken Refinement 


iamzoltan:
To address issues that did not fit into the scope of PR #214 and tick off the remaining desired functionality from #201

1) [x] Get trades history asynchronously using nursery and the count:
  - not sure we need this yet since realistically we'll be incrementally capturing trades data in the new ledger system. if we get to the point where user's are complaining about ledger loading on first boot, this might be something worth revisiting.
2) [x] ~Change the dict merging for trades to the new syntax: https://github.com/pikers/piker/pull/214/files#r830525470~ **this code was all rewritten thus obsoleting this change request**
3) [x] Get rid of the try except block in the positions calc: https://github.com/pikers/piker/pull/214/files#r830527334
4) [x] ~Change fixtures to accept kwargs: https://github.com/pikers/piker/pull/214/files#r817211683~ instead just expect `funtools.partial()`-ing in of needed state.

---
Further follow up TODOs (added by @goodboy):
- [x] ~we should probably use one of their [batch cancel endpoints](https://docs.kraken.com/rest/#operation/cancelAllOrders) in our [Client.submit_cancel()](https://github.com/pikers/piker/pull/214/files#diff-585ca6355c03f4dd8634e0d03e62bf791b8e60808a688027e5bd7b4af58d8c91R329)~ such that cc and other automated strats can destroy order sets faster. **-> turns out using the WS api makes this all fast enough to avoid needing the batch cancels (for now)**
  - [x] the other thing to investigate is whether the [ws order entry API](https://docs.kraken.com/websockets/#message-cancelOrder) doesn't have these same issues with throttling (which one would assume since it's not an extra http request per request). **-> it doesn't**

goodboy:
The one last bit was figuring out the source of latency in order cancellation.
I got a feeling it's possibly one of 2 things:
- their api is putting a speed bump (would be odd though)
- there's a `trio.sleep(1)` somewhere that isn't needed in our client

iamzoltan:
yea, the batch cancel endpoint I believe is the same as the single cancel one, only you pass in a list. when we use the `cc` command in piker, do we get all the cancels separately? or do they also come in a batch?

goodboy:
I've updated the issue descr with the new issues i found and probably have addressed / solved by now.

iamzoltan:
has all this is taking care of in the new branch?

goodboy:
[...]

Issue #297, opened 2022-04-09
`clearing._messages.BrokerdStatus` should not allow out-of-set statuses..


goodboy:
We have a [wrong comment saying there's an `'executed'` status](https://github.com/pikers/piker/blob/master/piker/clearing/_messages.py#L187) and should further maybe do some validation of the message at creation (though I'd rather wait until we just move to `msgspec.Struct`).

Main point is a user shouldn't be allowed to create a message that ems doesn't explicitly expect.

iamzoltan:
sounds good. is this something I should address in the same PR that addresses #296 or make a separate?

Issue #296, opened 2022-04-09
`kraken` backend: `BrokerdStatus.remaining` should decrement by fill amount


goodboy:
As per [comment in response to question](https://github.com/pikers/piker/pull/214#discussion_r846663818) in #214 *fill* msgs should report *how much* of the original order size was cleared in the transaction.

We probably just need a little arithmetic to finish this but I don't want to delay that PR any longer as many other things need to be built around/on it.

iamzoltan:
alright, I can tick this off next since it seems like it would be a quick PR.

iamzoltan:
did this also get sorted?

PR #298, opened 2022-04-09
Kraken cleaning

goodboy:
Follow up to #214 and includes i think at least fix for #290.

### DIFF
--- a/piker/brokers/kraken.py
+++ b/piker/brokers/kraken.py
@@ -33,7 +33,7 @@ import tractor
 from pydantic.dataclasses import dataclass
 from pydantic import BaseModel
 import wsproto
-from itertools import count
+import itertools
 import urllib.parse
 import hashlib
 import hmac
@@ -540,7 +540,8 @@ async def handle_order_requests(
 
     request_msg: dict
     order: BrokerdOrder
-    userref_counter = count()
+    userref_counter = itertools.count()
+
     async for request_msg in ems_order_stream:
         log.info(f'Received order request {request_msg}')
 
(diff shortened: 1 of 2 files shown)


piker/brokers/kraken.py
```diff
@@ -124,11 +124,11 @@ class Trade(BaseModel):
     Trade class that helps parse and validate ownTrades stream
 
     '''
-    reqid: str # kraken order transaction id
-    action: str # buy or sell
-    price: str # price of asset
-    size: str # vol of asset
-    broker_time: str # e.g GTC, GTD
+    reqid: str  # kraken order transaction id
```
goodboy, line 127: @iamzoltan fyi this is all the whitespace stuff you're missing bc no linter 😉

piker/brokers/kraken.py
```diff
@@ -616,42 +617,41 @@ async def handle_order_requests(
...
-            except AssertionError:
-                log.error(f'Order cancel was not successful')
+            # Check to make sure there was no error returned by
+            # the kraken endpoint. Assert one order was cancelled
+            assert resp['error'] == []
+            assert resp['result']['count'] == 1
+
+            # Check to make sure the cancellation is NOT pending,
+            # then send the confirmation to the ems order stream
+            pending = resp['result'].get('pending')
```
goodboy, line 627: please be sure this logic looks right, it's not clear to me at all why you assert things then catch the error around another block that has nothing to do with the assert. this is just generally non-pythonic to catch and handle asserts; the whole point is they raise.

piker/brokers/kraken.py
```diff
@@ -772,13 +756,21 @@ async def subscribe(ws: wsproto.WSConnection, token: str):
...
                             filled=float(trade.size),
                             reason='Order filled by kraken',
-                            # remaining='' # TODO: not sure what to do here.
                             broker_details={
                                 'name': 'kraken',
                                 'broker_time': trade.broker_time
-                            }
+                            },
+
+                            # TODO: figure out if kraken gives a count
```
iamzoltan, line 764: looks good so far.

I think kraken will send the individual trades made, even if the order was complete over several trades. does this mean we need to handle the decrement ourselves?

piker/clearing/_messages.py
```diff
@@ -184,7 +184,7 @@ class BrokerdStatus(BaseModel):
     # {
     #   'submitted',
     #   'cancelled',
-    #   'executed',
+    #   'filled',
```
iamzoltan, line 187: ah, so thats where I got it from haha. seems good now

goodboy: yah, made #297 as mentioned to addres this.

ideally this sorta thing isn't possible going forward.

please do test if this works if you can though 🏄🏼

iamzoltan: tested and works for live orders, not sure if it is also working for dark ones

piker/brokers/kraken.py
iamzoltan: no doubt. I need to get that up too haha

piker/brokers/kraken.py
iamzoltan: makes sense. Is this something you want to address in this PR?

goodboy: I don't really ever like seeing `except AssertionError:` unless in some kinda wonky testing (and even then).

If you think this works then i'm fine to land it - makes the code a lot more readable for me.

piker/brokers/kraken.py
goodboy: yeah i think so. doesn't have to be solved here which is why i made #296 to address it more specifically. unfortunately we'll likely have to test with real clearings?

piker/brokers/kraken.py
```diff
@@ -616,42 +617,41 @@ async def handle_order_requests(
...
-                            status='cancelled',
-                            reason='Order cancelled',
-                            broker_details={'name': 'kraken'}
-                        ).dict()
-                    )
-            except AssertionError:
-                log.error(f'Order cancel was not successful')
+            # Check to make sure there was no error returned by
+            # the kraken endpoint. Assert one order was cancelled
+            assert resp['error'] == []
```
goodboy, line 622: yeah like, normally the way i would do this is to try to get the result and then if one isn't there fail over to error lookup.

maybe i'll change this quick.

Issue #299, opened 2022-04-10
`kraken`: order mode modifcation of orders doesn't cancel/edit the order and instead just leaves them stale! 


goodboy:
Massive bug that i'm not sure how we missed...

Currently there's no use of the [`editOrder`](https://docs.kraken.com/rest/#operation/editOrder) endpoint and further no cancelling of submitted live orders such that if a user modified a live submission the original order **is left stale** without any indication of this to the user and further a new order is submitted at the new target price..

This is a massive issue and needs to be addressed asap!

I will be disabling kraken order mode support in #298 until this is resolved.

iamzoltan:
addressed in a new PR #300

iamzoltan:
PR #300 merged

PR #300, opened 2022-04-11
fix kraken bug, allow for live order edits

iamzoltan:
This seems to work. I tested it a bunch, but if @goodboy, @guilledk, or @wattygetlood could also test it out. That would be great.

I can't believe I missed this the first time hah.

### DIFF
--- a/piker/brokers/kraken.py
+++ b/piker/brokers/kraken.py
@@ -319,9 +319,9 @@ class Client:
         }
         if reqid is None:
             # Build order data for kraken api
-            data["ordertype"] = "limit"
-            data["type"] = action
-            data["volume"] = str(size)
+            data |= {
+                "ordertype": "limit", "type": action, "volume": str(size)
+            }
             return await self.endpoint('AddOrder', data)
         else:
             # Edit order data for kraken api
--- a/piker/brokers/kraken.py
+++ b/piker/brokers/kraken.py
@@ -33,7 +33,6 @@ import tractor
 from pydantic.dataclasses import dataclass
 from pydantic import BaseModel
 import wsproto
-import itertools
 import urllib.parse
 import hashlib
 import hmac
@@ -160,13 +159,6 @@ def get_config() -> dict[str, Any]:
 
     conf, path = config.load()
     section = conf.get('kraken')
-
-    if section:
-        log.warning(
-            'Kraken order mode is currently disabled due to bug!\n'
-            'See https://github.com/pikers/piker/issues/299'
-        )
-        return {}
 
     if section is None:
         log.warning(f'No config section found for kraken in {path}')
@@ -313,29 +305,41 @@ class Client:
         price: float,
         action: str,
         size: float,
-        reqid: int = None,
-    ) -> int:
+        reqid: str = None,
+    ) -> dict:
         '''
         Place an order and return integer request id provided by client.
 
         '''
-        # Build order data for kraken api
-        data = {
-            "userref": reqid,
-            "ordertype": "limit",
-            "type": action,
-            "volume": str(size),
-            "pair": symbol,
-            "price": str(price),
-            # set to True test AddOrder call without a real submission
-            "validate": False
-        }
-        return await self.endpoint('AddOrder', data)
+        if reqid is None:
+            # Build order data for kraken api
+            data = {
+                "ordertype": "limit",
+                "type": action,
+                "volume": str(size),
+                "pair": symbol,
+                "price": str(price),
+                # set to True test AddOrder call without a real submission
+                "validate": False
+            }
+            return await self.endpoint('AddOrder', data)
+        else:
+            # Edit order data for kraken api
+            data = {
+                "txid": reqid,
+                "pair": symbol,
+                "price": str(price),
+                # set to True test EditOrder call without a real submission
+                "validate": False
+            }
+            return await self.endpoint('EditOrder', data)
+
+
 
     async def submit_cancel(
         self,
         reqid: str,
-    ) -> None:
+    ) -> dict:
         '''
         Send cancel request for order id ``reqid``.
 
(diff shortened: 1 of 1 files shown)


iamzoltan:
this address #299, some CI tests fail, not sure why

goodboy:
CI is annoying new security stuff to do with git.

piker/brokers/kraken.py
```diff
@@ -313,29 +305,41 @@ async def submit_limit(
...
+                "validate": False
+            }
+            return await self.endpoint('AddOrder', data)
+        else:
+            # Edit order data for kraken api
+            data = {
+                "txid": reqid,
+                "pair": symbol,
+                "price": str(price),
+                # set to True test EditOrder call without a real submission
```
goodboy, line 332: Hm, can we expose this up through the method as kwarg too maybe?

Both endpoints support it so i think just a wholesale flag would be handy here if we try to implement a pseudo-paper engine no?

piker/brokers/kraken.py
```diff
@@ -313,29 +305,41 @@ async def submit_limit(
...
-            "volume": str(size),
-            "pair": symbol,
-            "price": str(price),
-            # set to True test AddOrder call without a real submission
-            "validate": False
-        }
-        return await self.endpoint('AddOrder', data)
+        if reqid is None:
+            # Build order data for kraken api
+            data = {
```
goodboy, line 316: the `data: dict` duplication isn't necessary right?

in the edit case can't you just update with the `txid` ?

goodboy (commented):
Left a couple quiffs but nothing deal breaking.

I'm going to test this now.

goodboy:
@iamzoltan oh only last thing would be disabling the config safeguard i put in yah?

piker/brokers/kraken.py
```diff
@@ -161,13 +160,6 @@ def get_config() -> dict[str, Any]:
     conf, path = config.load()
     section = conf.get('kraken')
 
-    if section:
```
goodboy, line 164: wait nm i missed this..

guilledk (approved):

piker/brokers/kraken.py
iamzoltan: what do you meam?

piker/brokers/kraken.py
iamzoltan: you mean, create the `dict` with the common keys from both end points, and then depending on which one we are using, add the extras?

piker/brokers/kraken.py
iamzoltan: not sure what you mean by this, can you elaborate?

piker/brokers/kraken.py
```diff
@@ -596,16 +597,21 @@ async def handle_order_requests(
...
                 )
             else:
-                # TODO: handle multiple cancels
+                # TODO: handle multiple orders (cancels?)
                 #       txid is an array of strings
-                reqid = resp['result']['txid'][0]
+                if order.reqid is None:
+                    reqid = resp['result']['txid'][0]
+                else:
+                    # update the internal pairing of oid to krakens
```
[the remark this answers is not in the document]
iamzoltan: basically, if its the first time the ems has seen the order, then the `reqid` is none, until we get a response from kraken. The the `txid` from kraken populates the `reqid` in the `ack` message.

then if the order is changed, and a `reqid` exists, the `EditOrder` endpoint gets pinged, and we need to change the `reqid` in the `ack` message to be associated with the new `txid` that is returned by kraken

piker/brokers/kraken.py
goodboy: the commented parameter that seems to make this a paper-like request?

piker/brokers/kraken.py
goodboy: yes

piker/brokers/kraken.py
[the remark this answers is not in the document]
goodboy: right so i think the answer to my question is just *yes*?

piker/brokers/kraken.py
goodboy: don't worry about it

piker/brokers/kraken.py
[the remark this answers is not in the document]
iamzoltan: yes?

piker/brokers/kraken.py
iamzoltan: aight

piker/brokers/kraken.py
iamzoltan: yes I hear you, as an arg in the `submit_limit` method, on it

iamzoltan (commented):
thanks for the comments @goodboy

ill get these pushes up asap

piker/brokers/kraken.py
```diff
@@ -308,34 +300,38 @@ async def get_trades(
...
             "pair": symbol,
             "price": str(price),
-            # set to True test AddOrder call without a real submission
-            "validate": False
+            "validate": validate
         }
-        return await self.endpoint('AddOrder', data)
+        if reqid is None:
+            # Build order data for kraken api
+            data["ordertype"] = "limit"
```
goodboy, line 322: fwiw this is a good spot to use the new merge syntax:

`data |= {'ordertype': 'limit', 'type': action, 'volume': str(size)}`

if you wanted but nbd and it's def not a requirement for this pr.

goodboy: oh and i also prefer the name `payload` for inputs to endpoints just for the next time.

goodboy (approved):
Good enough for me and I tested and everything lookin right.

Nice work @iamzoltan 🏄🏼

piker/brokers/kraken.py
iamzoltan: oh slick, I like that new syntax, I will update it. I can also switch data to payload if that is consistent with the rest of the base

iamzoltan (commented):
gonna update that syntax, test it out, push and then merge. Thanks for the comments

piker/brokers/kraken.py
iamzoltan: acutally given ill leave the payload change for next time. <3

PR #304, opened 2022-04-15
Offline history loading

goodboy:
Adds `ib` offline (market is currently closed - rt quotes aren't happening) history loading.
This brings in a bunch of commits factored out of #289 and some manual work for windows.

Please feel free to test for those users that are on this backend.

---
Oh, right because we depended on it, the factored commits also included display loop re-works from #289 which allows manually triggering and update cycle with the original motivation being for historical prepends when the market is not open (so updates from rt quotes aren't triggering updates).

> factor the "graphics draw cycle" steps into a sync function that can be triggered (remotely) to update on certain types of events for example when an fsp running in a subactor makes an update but the market is not open - it needs to tell the chart app to update graphics

### DIFF
--- a/piker/ui/_display.py
+++ b/piker/ui/_display.py
@@ -290,6 +290,7 @@ async def graphics_update_loop(
 def graphics_update_cycle(
     ds: DisplayState,
     wap_in_history: bool = False,
+    trigger_all: bool = False,  # flag used by prepend history updates
 
 ) -> None:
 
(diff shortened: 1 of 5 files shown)


piker/ui/_display.py
```diff
@@ -109,6 +111,33 @@ def chart_maxmin(
     return last_bars_range, mx, max(mn, 0), mx_vlm_in_view
 
 
+@dataclass
+class DisplayState:
```
guilledk, line 115: Love this new `dataclass`

guilledk (approved):

piker/ui/_display.py
goodboy: Yeah makes it super nice to manual updates - probably will work for remote control decent down the road as well.

piker/ui/_display.py
```diff
@@ -231,222 +281,263 @@ async def graphics_update_loop(
...
+
+def graphics_update_cycle(
+    ds: DisplayState,
+    wap_in_history: bool = False,
+    trigger_all: bool = False,  # flag used by prepend history updates
+
+) -> None:
 
+    # TODO: eventually optimize this whole graphics stack with ``numba``
+    # hopefully XD
```
iamzoltan, line 296: heh, I hope so

iamzoltan (approved):
looks great, can't wait to get something like this for good ol' phil

PR #305, opened 2022-04-16
Crypto$ bars: lotsa 1m for ze pumpbois

goodboy:
This is like **super duper duper** WIP / draft but should be a gateway to crypto-bois for mucking with large data sets to check the performance of stuff brought in by #289 and #247..


goodboy:
Replaced by #308.

PR #302, opened 2022-04-14
Incremental update paths

goodboy:
WIP building out a "flows" layer for incremental update and multi-actor graphics rendering.

Issue #90, opened 2020-05-16
Choosing a time series database


goodboy:
Did some decent starter research on what systems will work for production grade data feed collaboration.

I'm going to list some options and what I deem to be pros/cons.

- alpaca's [marketstore](https://github.com/alpacahq/marketstore)
  - built [for use with `numpy`](https://github.com/alpacahq/pymarketstore/tree/81841c95a7f95d7fe9d2e5eeb3fa9dc5a85557df#examples)
  - it seems simple, with a reasonably manageable code base in golang; hackable
  - recently got [grpc support](https://github.com/alpacahq/marketstore/pull/316)
  - [plugin system](https://github.com/alpacahq/marketstore/tree/master/plugins) which seems to a a few decent things you might not find in your average TSDB:
    - a websocket [streaming interace](https://github.com/alpacahq/marketstore/tree/master/contrib/stream) on updates
    - an [on disk aggregator](https://github.com/alpacahq/marketstore/tree/master/contrib/ondiskagg)
  - no clustering, failover, sharding
- [influxdb](https://en.wikipedia.org/wiki/InfluxDB) is the gold standard in TSDBs
  - here's a [quan related SO question](https://quant.stackexchange.com/questions/20625/performance-of-open-source-time-series-database-for-financial-market-data) that resulted in [this pump](https://www.algotrader.com/algotrader-3-1-integrates-influxdb/)
- [timescaledb](https://github.com/timescale/timescaledb) is an extension on top of postgres
  - here's a 18' pump on [pushing 3B points a day](https://www.dnsfilter.com/blog/why-dnsfilter-replaced-influxdb-with-timescaledb/) and how it outperforms influxdb
  - another [pump blog post](https://streamsets.com/blog/ingesting-data-from-apache-kafka-to-timescaledb/) on ingesting from `kafka`

Additionally,
- [blog pos](https://arrow.apache.org/blog/2017/06/16/turbodbc-arrow/)t on connecting apache arrow to the SQL world
- a post from uber on [moving from postgres to mysql](https://eng.uber.com/postgres-to-mysql-migration/)

One of the things I want to keep in mind is supporting [apache arrow](https://arrow.apache.org/blog/) formatted data for streaming.
The arrow project's new [flight](https://arrow.apache.org/blog/2019/10/13/introducing-arrow-flight/) system seems to be the way things are going.

goodboy:
Probably worth reading up on some professional grade systems:
[...]

PR #308, opened 2022-05-08
Storage layer: initial `marketstore` tsdb support with async OHLCV history loading.

goodboy:
Replaces #247 and #305 (merging history from both) and instead adds a more formal "storage layer" for retrieving and storing large ohlcv series from all major backends. The work here is best experienced with the new incremental update patchset from #302 in order to see the new graphics performance improvements at work though this should work moderately well as is.

---
#### TL;DR:
- adds a `marketstore` docker container supervisor actor which allows spawning by passing `pikerd --tsdb` -> this closes https://github.com/pikers/piker/issues/143
  - this command **must** be run as root, but the root perms will be dropped (in `pikerd`) shortly after `marketstored` (the container super) is started
- adds working `async with open_history_client()` support to `ib`, `binance`, `kraken` which allows pulling in and storing a large amount of history in the tsdb, `marketstore`.
  - this API expects the async cntx mngr endpoint to deliver an awaitable that can be called with an input `datetime` range, new special exception signals can be used (`NoData`, `DataUnavailable`) to signal that the backend can't deliver data for a given range (or at all). 
- async "batch" fetching of OHLC history using `trimeter` in the backends that can handle it (currently only `binance`).
  - includes a history "frame generator" system which delivers `datetime` ranges that are passed to the `trimeter` request scheduler and which dynamically adjusts the request-time index when a gap is detected.
- adds a `piker.data.marketstore.Storage` api/layer which allows async, high level operations. the intention is to eventually have this layer support more tsdb providers like `arctic`, `techtonicdb`. Currently the only backend is `marketstore` with client-side operations implemented using our [`anyio-marketstore` library](https://github.com/pikers/anyio-marketstore):
  - loading / reading existing ohlc time series by `fqsn` with appropriate request size limiting with `.read_ohlcv()`
  - writing ohlcv series by `fqsn` with appropriate limits with `.write_ohlcv()`
  - deleting time series entries via `Storage.delete_ts()`

---
#### What this **does (yet) not** introduce:
- real-time ingest from tick feeds to `marketstore`
  - this was originally planned but the shorter path to get graphics downsampling methods up and running was to first start with OHLC history ingest and display
- a full `storesh` interactive repl for managing the tsdb, there is a minimal `ipython` embedding at the moment but it is nowhere complete and we need a follow up task-issue to finish this.

---
#### TODO:
- `._ahab.py` supervisor:
  - [x] probably deliver back net-socket info over the `ctx.started()` call?
    - [x] grprc socket
    - [x] ws addr
  - [x] figure out what to do with the `mkts.yaml` config?
  - [x] cli support for running `pikerd` with the tsdb stuff spawned?
    - [x] maybe a `--tsdb` or `--data` or something?
  - [x] step by step `pikerd --tsdb` test list:
    - [x] `pikerd --tsdb` should raise `DockerNotStarted` and appropriate perms error on no `sudo` (for now)
    - [x] ctrl-c should kill container instance
- general `marketstore` config and operation:
  - [x] should we always push newly received history from backends which is not yet in the tsdb to it?
    - we could also just offer this as a config option? eventually the UI should offer manual controls for such things..
  
 to be done as #313 
  - [ ] what docs should we offer regarding saving / deleting history?

---    
#### follow up (to be written in new task-issues and implemented in coming PRs)

 moved to #314
- [ ] tick ingest support and an accompanying feed-style inter-actor API to pull feeds from ingestor re-broadcast system(s):
  - [ ] tick ingest to `marketstore` from `brokerd` feeds and experiment with `techtonicdb` schema (some tinkering was already done in this patchset by @guilledk but is unfinished).
  - [ ] tick-to-ohlcv sampling if it can be done with the [aggregator plugin](https://github.com/alpacahq/marketstore/tree/master/contrib/ondiskagg) (got a feeling we'll need to write at least a `1Sec` bucket in order for this to work looking at the code:
      - [`aggregate()` calls into `model.FromTrades()`](https://github.com/alpacahq/marketstore/pull/399/files#diff-392952832ca30145443aff46786de412b7b5876a33e5dbcbb34dde20637c6370R285) and [`FromTrades()` only accepts a min step of `1Sec`](https://github.com/alpacahq/marketstore/pull/399/files#diff-cf7320bc8ecabe28fa6f7712d8df43143d3fa8577667f777c792db773ea5fbdbR192-R205)
      - we might be able to either write a golang plugin/extension to do it from pure tick data or we can make this part of a piker actor?

these moved to https://github.com/pikers/piker/issues/312
- [ ] we need a timeseries diffing and syncing system to validate new captured histories from real-time runs (which almost always are slightly different then the history provided by the providers dbs 🙄) as well as for catching history mis-writes / gaps which need to be edited / corrected when bugs
- [ ] a REPL (with `ipython`) that allows interaction, edit, and general management of the tsdb for both the purposes of research and just plain old data mgmt.

### DIFF
--- a/piker/data/_ahab.py
+++ b/piker/data/_ahab.py
@@ -30,7 +30,7 @@ from trio_typing import TaskStatus
 import tractor
 import docker
 import json
-from docker.models.containers import Container
+from docker.models.containers import Container as DockerContainer
 from docker.errors import DockerException, APIError
 from requests.exceptions import ConnectionError, ReadTimeout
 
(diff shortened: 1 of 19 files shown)


piker/data/cli.py
```diff
@@ -49,51 +57,58 @@
...
-                for sym in names:
-                    # tbk = _tick_tbk.format(sym)
-                    tbk = tuple(sym, *_tick_tbk_ids)
-                    print(f"Destroying {tbk}..")
-                    await client.destroy(mk_tbk(tbk))
-            else:
-                print("Nothing deleted.")
 
-    tractor.run(main)
+# @cli.command()
```
goodboy, line 78: ahh yeah we can probably drop this.

the way i'd like to deal with "tsdb management" is the new storesh repl (with embedded ipython) which will have a small interactive API for manual db tinkering.

piker/data/_source.py
```diff
@@ -254,61 +254,6 @@ def iterfqsns(self) -> list[str]:
         return keys
 
 
-def from_df(
```
goodboy, line 257: ahh yeah right!

we can drop this (and maybe pandas altogether for now) since i rewrote the ib ohlc frame parser to just cast directly to `numpy`.

piker/data/_ahab.py
```diff
@@ -0,0 +1,422 @@
...
+    task_status: TaskStatus[trio.Event] = trio.TASK_STATUS_IGNORED,
+
+) -> None:
+    '''
+    Start a ``docker`` container supervisor with given service name.
+
+    Currently the actor calling this task should normally be started
+    with root permissions (until we decide to use something that doesn't
+    require this, like docker's rootless mode or some wrapper project) but
+    te root perms are de-escalated after the docker supervisor sub-actor
```
goodboy, line 362: s/te/the

piker/fsp/_volume.py
```diff
@@ -309,7 +309,7 @@ async def flow_rates(
 
         if period > 1:
             trade_rate_wma = _wma(
-                dvlm_shm.array['trade_count'],
+                dvlm_shm.array['trade_count'][-period:],
```
goodboy, line 312: Oh right, this was a totally unrelated patch to improve FSP actor perf since the calc prior was recomputing from all history, which once you have a shite ton of data loaded is pretty costly 😂

piker/data/marketstore.py
```diff
@@ -123,146 +169,458 @@ def quote_to_marketstore_structarray(
...
+    60*5: '5Min',
+    60*15: '15Min',
+    60*30: '30Min',
+    60*60: '1H',
+    60*60*24: '1D',
+})
+
 
-    async def destroy(
+class Storage:
```
goodboy, line 226: Oh right, so I want to probably move this layer to a new module (`._storage.py` or whatever) and we can start to plan for supporting more backends if and where we run into problems or find better alternatives.

See https://github.com/pikers/piker/issues/90 for a list of alts on the horizon.

guilledk (approved):

goodboy:
Pushed a few more `piker.data._sharedmem` changes from #302 which fixed some shm pushing / slicing edge cases.

piker/data/cli.py
goodboy: actually gonna leave it commented in here just in case we decide std cli cmds for this is handy.

piker/data/marketstore.py
goodboy: #315

Issue #307, opened 2022-05-05
`ib` uses "FIFO" positioning logic - that's super anti-retail.. lame


goodboy:
i got lots more examples of plain old pure nonsense that goes on with this broker (and hint, issues like this **are exactly** why this project got started by a bunch of weirdos in the first place) but the latest is a truly exemplary *inverse-hanlon's*, icing-on-cake, can you be more rigged, idiotic, plain-laughable-while-despicable.

The easiest way to understand the issue in it's entirety is to read [this elitetrader thread](https://www.elitetrader.com/et/threads/fifo-rule-kills-futures-traders-profits.352534/) which breaks down the issue in great detail.

### synopsis:
`ib`'s stack, despite being probably the most advanced in the retail space, *magically* can't give you more then **7days** worth of your own trade events without using their even shoddier "PLEX RePOrtz" (which you then have to parse and correlate on your own with the traditional APIs, oh and did i mention it's delayed a day...`smh**2`); even the worst crypto exchanges give you at least your trades history ledger (even if they don't compute pps or PnL for you 😿 ). 

I'd really love to hear the explanation for **why** literally the most important ledger as a trader isn't available through the API over a timeframe suitable for actually calculating long running pp pnls..

**What `ib` will give you** is their horribly calculated FIFO style "average price" in PnL events and by default in TWS's (mosaic) "portfolio view".  FIFO is horrible because as you **exit a pp** your **average price changes**. This is just plain idiotic: it means the trader has to manually recompute what the "magic price is" that determines whether you exit the pp at a profit or a loss (sorry to anyone who can't grok this but i have no clue how you can even compute a net-zero pnl without this).  Using FIFO is mysteriously convenient for the broker (and i'm sure their larger clients who trade against most retail traders who get stuck with this accounting reporting model) yet in pretty much all the ways, makes their client's more susceptible to lose money unless they're saavy enough to compute a LIFO style pp avg price themselves. I'll say it again, i don't know how you can calculate a sane pnl with LIFO - i've experienced it first hand - the avg price will **never be** the price that determines the net-zero position pnl.

### technicalities
 - ib's ["PnL Eventz"](https://interactivebrokers.github.io/tws-api/pnl.html) i've now first hand experienced, do FIFO "pnl calculations" and made me take a larger loss (on a stop out) then intended since the "average price" calculated from a now *scaled out of positiion* was "lower" then it should have been..
[...]

PR #321, opened 2022-05-15
Ib dedicated data client

goodboy:
Adds a new policy for the `ib` backend which insists the user must choose a priority for a "data client' connection: a specific (set of) client(s) which is anticipated to be used for data feeds, history loading, and symbol search as opposed to **just for orders / trading**.

The user **must now add** a `brokers.toml` section which determines the data client priority by order of account name:
```toml
# when clients are being scanned this determines
# which clients are preferred to be used for data
# feeds based on the order of account names, if
# detected as active on an API client.
prefer_data_account = [
    'paper',
    'margin',
    'ira',
]
```

The main motivation for this is to eventually allow users to run less sensitive API gateways on remote hosts which do data work whilst keeping more sensitive live gateway nodes local (or on diff hosts). This also paves the way for being able to do targetted data resets without resetting gws which are only doing orders. Further this is just more specific about which API node is doing what.

----
Other tidbits:
- added `Client.trades()` to pull executions/fills from the current session to begin working towards #307
- use the newer `MethodProxy` stuff for `asyncio` integration throughout all endpoints and drop the old *one-task-per-method-call* style and helpers


guilledk (approved):

config/brokers.toml
```diff
@@ -14,14 +14,26 @@ secret = ""
...
+prefer_data_account = [
+    'paper',
+    'margin',
+    'ira',
+]
+
+# the order in which ports will be scanned
+# (by the `brokerd` daemon-actor)
+# is determined # by the line order here.
 ports.gw = 4002
```
goodboy, line 30: Got a feeling this is still broken when scanning for more then one port # (can't member why tho) so probably needs at least some manual testing before merge..

goodboy: Well testing says otherwise once I tweaked a couple things 😂 

At least one more commit coming, i might also toss in a commit that changes the feed status label thing on the sidepane to show the current "data client" socket info.

piker/brokers/ib.py
```diff
@@ -1858,122 +1952,132 @@ async def stream_quotes(
...
+            # TODO: for loop through all symbols passed in
+            init_msgs = {
+                # pass back token, and bool, signalling if we're the writer
+                # and that history has been written
+                sym: {
+                    'symbol_info': syminfo,
+                    'fqsn': first_quote['fqsn'],
+                },
+                'status': {
+                    'data_ep': f'{host}:{port}',
```
goodboy, line 2000: Shows the preferred data client sockaddr in the feed status label in the sidepane.

PR #322, opened 2022-05-21
Dockerize `ib-gw` 🏄🏼 

goodboy:
Based on the now defunct project @
https://github.com/waytrade/ib-gateway-docker

Adds a `docker-compose.yml` and necessary gateway and `IBC` config
files to make it possible to spin up a local gateway on localhost:4002
and connect to it without issue using `ib_insync`.

---
TODO: container POC config
- [ ] ~add a api socket blackhole for anything but connections from localhost~ **bleh we can't; no `iproute2` installed in this image...**:
  - `ip rule add blackhole iif <docker_net_iface_name> to 0.0.0.0/0 dport 4002`
  - `ip rule add not unicast iif lo to 0.0.0.0/0 dport 4002` **actually this is probably better since it's the inverse and requires no net iface info**
- [ ] figure out why the vnc auth still doesn't seem to be working?
- [x] clarify the vnc port setting (**explained in readme from [48a8174](https://github.com/pikers/piker/pull/322/commits/48a817470e068dcac15b148eeac831a9d3667d18)**)

---
In a follow up PR, we'll want to,
- automated the equivalent docker-compose steps using our
  `.data._ahab` supervisor system
- probably simplify and roll our own container (likely alpine or nixos
  based) which drops uneeded deps (`socat`, vnc) and adds `xdotool`.
- allow for API socket mapping to just be pulled direct from
  a user's `brokers.toml` and we'll just pass that direct to `IBC`'s
  config.
  
---
#### Follow up TODOs for windows:
For extra lolz, you can't use `--net=host` on windows as per:
- https://docs.docker.com/network/host/
   > The host networking driver only works on Linux hosts, and is not supported on Docker Desktop for Mac, Docker Desktop for Windows, or Docker EE for Windows Server.
   
So that means we'll need to:
- [ ] either go back to the `socat` stuff from the original container from waytrade
- [ ] use a `bridge` network inside the container namespace and do sensible routing table to map the necessary vnc and gw ports

I'm thinking latter is better and we can just configure this dynamically if/when we do container running from our `._ahab.py` supervisor daemon.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -138,7 +138,7 @@ provider support
 ****************
 for live data feeds the in-progress set of supported brokers is:
 
-- IB_ via ``ib_insync``
+- IB_ via ``ib_insync``, also see our `container docs`_
 - binance_ and kraken_ for crypto over their public websocket API
 - questrade_ (ish) which comes with effectively free L1
 
@@ -150,6 +150,7 @@ coming soon...
 if you want your broker supported and they have an API let us know.
 
 .. _IB: https://interactivebrokers.github.io/tws-api/index.html
+.. _container docs: https://github.com/pikers/piker/tree/master/dockering/ib
 .. _questrade: https://www.questrade.com/api/documentation
 .. _kraken: https://www.kraken.com/features/api#public-market-data
 .. _binance: https://github.com/pikers/piker/pull/182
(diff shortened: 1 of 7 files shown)


guilledk (approved):

goodboy:
All other unchecked tasks moved to #329 FYI

PR #334, opened 2022-06-06
`pikers.broker.ib` subpackage

goodboy:
A conversion of `piker.brokers.ib.py` into a sub-package with the following new modules:
- `ib.api.py` containing the core `ib_insync` wrapper `Client` and low level `trio`-`asyncio` wrapping / translation layer using `tractor.to_asyncio`
- `ib.feed.py` which contains all data related endpoints for historical and real time feeds
- `ib.broker.py` which contains the required trade endpoint(s) needed to integrate with `piker`'s EMS~


goodboy:
it's actually #335

Issue #325, opened 2022-06-04
Faster y-range sorting


goodboy:
After #302 lands there are 2 main sources of latency left in chart interaction:
- redraw latency (a much larger task to deal with that will require further extensions to our `Render`ing sub-systems)  that contains:
  - maybe a downsampling of data via m4
  - a delete and redraw of a `QPainterPath` for the data in view
  
- `max()` /  `min()` sorting of the y-data *in-view*
  - currently [we naively use `np.max()`/`.min()`](https://stackoverflow.com/questions/64503589/what-is-the-fastest-way-to-find-the-min-and-max-value-of-a-large-multidimensiona) and [very basic x-range caching inside our `Flow` type](https://github.com/pikers/piker/blob/ab135d2b7578f163d66c1af4c0c7d3ead8936aa2/piker/ui/_flows.py#L375)
  - the display loop will also naively call this same method on updates, which if there's conditions to actually redraw a large (1M datums +) entire graphic in view, can be quite slow
  
In order to see this latency in action pass a `--profile 10` to your `piker chart` command to see something like:

```python
 > Entering `ChartView._set_yrange()`: `dolla_vlm`
    callback ._maxmin(): (0, 22800345.0): 24.8241 ms
    set limits: (-1368020.7, 24168365.7): 0.1694 ms
  < Exiting `ChartView._set_yrange()`: `dolla_vlm`, total time: 24.9964 ms
    > Entering `<piker.ui._chart.ChartPlotWidget object at 0x7f8b66fcae60>.maxmin(name=trade_rate)`: `volume`
      volume got bars range: 0.0380 ms
      yrange mxmn: (0, 1299940) -> (0.0, 0.0): 16.0576 ms
    < Exiting `<piker.ui._chart.ChartPlotWidget object at 0x7f8b66fcae60>.maxmin(name=trade_rate)`: `volume`, total time: 16.1166 ms
    > Entering `<piker.ui._chart.ChartPlotWidget object at 0x7f8b66fcae60>.maxmin(name=dark_trade_rate)`: `volume`
      volume got bars range: 0.1704 ms
      yrange mxmn: (0, 1299940) -> (0.0, 0.0): 12.8711 ms
    < Exiting `<piker.ui._chart.ChartPlotWidget object at 0x7f8b66fcae60>.maxmin(name=dark_trade_rate)`: `volume`, total time: 13.0730 ms
  > Entering `ChartView._set_yrange()`: `trade_rates`
[...]

PR #342, opened 2022-06-26
Mxmn from m4

goodboy:
~Draft~ **Working POC** for an idea I had while pondering #325.

The idea is simple: pull total input yrange max and min from the m4 output since it's already computing the values per pixel window and we just need to do one more max min each step and then return the summary values with the output.

---
#### MAIN THINGS TO TEST:
- [x] pop a chart on this branch **and** one that is based on master's code ideally on the same symbol with more then 2 years worth of 1m OHLC
- [x] check if one feels faster then the other
- [x] do the same with an added `--profile 10` to the `chart` cmd and see if you notice any further difference and/or worse numbers from the actual profiler output

---
Notes:
- [x]  ~right now this seems to be most effective on long term crypto (1m) charts and slower (read not as fast) on 1s feeds (due to the large zeroed arrays allocated for trade rates / dark vlm?)~ now fixed via https://github.com/pikers/piker/pull/342/commits/c6205175435fd03c90de87801abf4021ef4c847a which means even zero-ed series are cached correctly and we get the same during-downsample-mxmn speedup effects 😎  
- obviously the mxmn summary valules are computed with a flow is **not** being downsampled so there's still room for improvement presuming using `numba` or a streaming algo (like demire see #325) would do better for zoomed in / source data graphics in view cases.
- the whole `.maxmin()` callback flow i still find super confusing, mostly because we have fsps on subcharts which each have flows but which also are somewhat needlessly calling `ChartPlotWidget.maxmin()` for each overlayed flow in a `PlotItem`
  - ideally we can somehow simplify this whole `ChartView._maxmin()`, `._set_yrange()`, `Flow.maxmin()` orchestration to be more "top level triggered" by whatever view is in use and then not have Qt signal callbacks firing all over the place? needs more thought and definitely a rework to make it more understandable (unfortunately this is growing pains of figuring out how to make Qt's signals wackiness fit with our async engine..)
  - [x] made a follow up issue to begin digging into solutions for this mess in  #343

### DIFF
--- a/piker/ui/_display.py
+++ b/piker/ui/_display.py
@@ -226,32 +226,8 @@ async def graphics_update_loop(
     tick_margin = 3 * tick_size
 
     chart.show()
-    # view = chart.view
     last_quote = time.time()
     i_last = ohlcv.index
-
-    # async def iter_drain_quotes():
-    #     # NOTE: all code below this loop is expected to be synchronous
-    #     # and thus draw instructions are not picked up jntil the next
-    #     # wait / iteration.
-    #     async for quotes in stream:
-    #         while True:
-    #             try:
-    #                 moar = stream.receive_nowait()
-    #             except trio.WouldBlock:
-    #                 yield quotes
-    #                 break
-    #             else:
-    #                 for sym, quote in moar.items():
-    #                     ticks_frame = quote.get('ticks')
-    #                     if ticks_frame:
-    #                         quotes[sym].setdefault(
-    #                             'ticks', []).extend(ticks_frame)
-    #                     print('pulled extra')
-
-    #                 yield quotes
-
-    # async for quotes in iter_drain_quotes():
 
     ds = linked.display_state = DisplayState(**{
         'quotes': {},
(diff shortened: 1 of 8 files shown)


piker/ui/_display.py
```diff
@@ -416,10 +420,8 @@ def graphics_update_cycle(
             )
             or trigger_all
         ):
-            # TODO: we should track and compute whether the last
-            # pixel in a curve should show new data based on uppx
-            # and then iff update curves and shift?
             chart.increment_view(steps=i_diff)
+            # chart.increment_view(steps=i_diff + round(append_diff - uppx))
```
goodboy, line 424: left this is in because to me the logic makes sense but for whatever reason it doesn't seem to actually be correct based on testing.

Ideally we can get to where there's never an error margin on shifts when optimizing for uppx scaling.

piker/ui/_flows.py
```diff
@@ -416,6 +417,10 @@ def maxmin(
             if not slice_view.size:
                 mxmn = None
 
+            elif self.yrange:
```
goodboy, line 420: the bit where we choose to use the m4 yrange output if set instead of the normal "manual" sorting calls from prior.

guilledk (approved):

piker/ui/_fsp.py
```diff
@@ -776,12 +781,16 @@ def chart_curves(
 
             ) -> None:
                 for name in names:
+
+                    render = False
```
goodboy, line 785: ahh this can be dropped..

PR #346, opened 2022-06-29
Kraken ledger pps

goodboy:
`piker.brokers.kraken` pkg reorg into multiple submods (l like #334 for `ib`) as well as initial ledger file pull and update support and use of our new `pps.toml` system from #336.

~Putting this up as a draft to start until the pps stuff is written.~ now complete!

Probably will *not* be solving #290 related bugs until a follow up PR but this should give a good base to work on those.

This **does solve** #310's pp updates part (which was entirely missing, we were only getting pp updates on restarts of the backend..) **and** #311.

PR #336, opened 2022-06-07
Lifo pps for `ib`

goodboy:
Proper attempt at #307 and a local "trades" ledger system using `.toml` files under `<confdir>/ledgers/` where files will be named something like: `trades_<brokername>_<accountname>.toml`

So far ledger contents looking something like this:
<details>
  <summary>Click to expand!</summary>
 
```toml
[ib.20220512.977661167]
accountId = "DUDUDUDUD"
acctAlias = ""
model = ""
currency = "USD"
fxRateToBase = 1.3046
assetCategory = "FUT"
symbol = "MGCM2"
description = "MGC 28JUN22"
conid = 430360671
securityID = ""
securityIDType = ""
cusip = ""
isin = ""
listingExchange = "NYMEX"
underlyingConid = 79702479
underlyingSymbol = "MGC"
underlyingSecurityID = ""
underlyingListingExchange = ""
issuer = ""
multiplier = 10
strike = ""
expiry = 20220628
tradeID = 977661167
putCall = ""
reportDate = 20220512
principalAdjustFactor = ""
dateTime = "20220512;115003"
tradeDate = 20220512
settleDateTarget = 20220513
transactionType = "ExchTrade"
exchange = "NYMEX"
quantity = 1
tradePrice = 1834.3
tradeMoney = 18343
proceeds = -18343
taxes = 0
ibCommission = -0.77
ibCommissionCurrency = "USD"
netCash = -97.77
closePrice = 1824.6
openCloseIndicator = "O"
notes = ""
cost = 18343.77
fifoPnlRealized = 0
fxPnl = 0
mtmPnl = -97
origTradePrice = 0
origTradeDate = ""
origTradeID = ""
origOrderID = 0
clearingFirmID = ""
transactionID = 3896422621
buySell = "BUY"
ibOrderID = 398090454
ibExecID = "0000e1a7.627cece5.01.01"
brokerageOrderID = "00dd230498015d2347c93b9.0001"
orderReference = ""
volatilityOrderLink = ""
exchOrderId = "00d345980012a0234dd49[2]"
extExecID = "237147708B"
orderTime = "20220512;115003"
openDateTime = ""
holdingPeriodDateTime = ""
whenRealized = ""
whenReopened = ""
levelOfDetail = "EXECUTION"
changeInPrice = 0
changeInQuantity = 0
orderType = "LMT"
traderID = ""
isAPIOrder = "Y"
accruedInt = 0
serialNumber = ""
deliveryType = ""
commodityType = ""
[...]

PR #347, opened 2022-07-02
Pps postmortem

goodboy:
Repairs from #336 mostly in the `Position.minimize_clears()` logic which was previously incorrect: the arithmetic needed to be inverted and the final zero-ing event included.

Some further tiny cleanups in the `ib` backend code.
Oh, and a change to the ems's position caching table / relaying to prep for #346.

### DIFF
--- a/piker/clearing/_ems.py
+++ b/piker/clearing/_ems.py
@@ -289,7 +289,11 @@ class TradesRelay:
     brokerd_dialogue: tractor.MsgStream
 
     # map of symbols to dicts of accounts to pp msgs
-    positions: dict[str, dict[str, BrokerdPosition]]
+    positions: dict[
+        # brokername, acctid
+        tuple[str, str],
+        list[BrokerdPosition],
+    ]
 
     # allowed account names
     accounts: tuple[str]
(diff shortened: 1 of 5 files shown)


piker/clearing/_ems.py
```diff
@@ -578,11 +588,9 @@ async def translate_and_relay_brokerd_events(
 
             relay.positions.setdefault(
                 # NOTE: translate to a FQSN!
-                f'{sym}.{broker}',
-                {}
-            ).setdefault(
-                pos_msg['account'], {}
-            ).update(pos_msg)
+                (broker, sym),
```
goodboy, line 591: So instead of a table of tables we just do a flat table with a `tuple[str, str]` of the broker and account name.

piker/pp.py
```diff
@@ -285,15 +285,14 @@ def minimize_clears(
         all transactions before the last net zero size to avoid
         unecessary history irrelevant to the current pp state.
 
-
         '''
-        size: float = 0
+        size: float = self.size
```
goodboy, line 289: This and the subtract change below were the main issue with getting the `pps.toml` clears tables per position correct..

guilledk (approved):

guilledk (approved):

Issue #345, opened 2022-06-29
Positioning system refinements: paper engine, file writing, toml style and perf 


goodboy:
After #336 lands we'll have a first draft of a completely
broker-agnostic position tracking system that **doesn't use the classic
FIFO style** and further allows to offer custom dynamic pricing models
depending on user desire.

---
As a recap #336 introduces:
- APIs for writing *trades ledgers* per backend broker
  - the initial idea is to allow writing "native" schema-ed/formatted trade entries to a `.toml` file and then expect the respective backend to provide (a) routine(s) to normalize to a common format for later `piker` processing into,
- a new config-ledger file: `pps.toml` which is an incrementally updated data file containing the minimal set of information to allow a trader to immediately be able to reproduce any position for an instrument they are trading:
  - here's an example for micro-daq futes take from #336 :
```toml
[ib.algopaper."mnq.globex.20220916"]
size = 58.0
be_price = 11703.458965517251
bsuid = 497954580
expiry = "2022-09-16T00:00:00+00:00"
clears = [
 { tid = "0000e1a7.629f3c2d.01.01", cost = 0.57 },
 { tid = "0000e1a7.629fa5c0.01.01", cost = 0.57 },
 { tid = "0000e1a7.62a0a898.01.01", cost = 0.57 },
 { tid = "0000e1a7.62a0fb24.01.01", cost = 0.57 },
]
```

---
### Desired outstanding features: (some of these might get broken into new issues)
|-> most ledger content moved to #510 B)

- correctness improvements, API ergonomics: **=> moved to #510**
- audit system **moved to #510**

---
#### paper engine support (**more or less complete**)

- [x] support for the paper engine such that users can forward test
  strats and keep progress in `git` and have a detailed clear-by-clear
  record of algo behaviour
  - [x] we need to figure out a minimal trades ledger schema probably
    based on an amalgamation of the best parts of other backends
    (**prolly landed in https://github.com/pikers/piker/pull/462 but who
    knows**)
 => **more or less determined by `Transaction` type**

---
#### CLI tools:
**make a new issue for this and stuff from #489**
- CLI interface for all this: see #512

---
#### novel UI ideas: 
**make new issue for this and stuff from #489**
- [x] historical position *line graphics* on charts which show breakeven
  price changes overlayed with actual underlyings, aka *pp liftetime
  "lines"* more or less as super simple FSP style graphics that show the
  beginning, changes, and end to a given pp such that users can easily
  see past trades and their state changes on top of historical data
[...]

Issue #310, opened 2022-05-09
`kraken` **live and dark** orders don't send correct fill messages,  pp doesn't update... etc.


goodboy:
just entered via one and none of the following worked in the chart order mode UI:
- no fill message received and thus no fill triangle annotation placed on the chart
- no position update and thus no pp line
- no notification of a fill anywhere..

Looking at these logs i dunno why exactly the UI isn't updating (since the order did clear) but it's likely a messaging or `reqid` mismatch i'm guessing.

Probably also relates to work described in #290.
 
#### `pikerd` logs:
```python

May 09 13:41:19 (emsd, 34778, _emsd_main)) [INFO] piker.clearing._ems _ems.py:787 Received order cmd:
{'account': 'kraken.spot',
 'action': 'buy',
 'brokers': ['kraken'],
 'exec_mode': 'dark',
 'oid': '2169fa61-b143-40cf-984a-3eb6d80bf213',
 'price': 29827.0,
 'size': 0.00419083,
 'symbol': 'xbteur.kraken'}
 
May 09 13:53:07 (emsd, 34778, piker.clearing._ems.clear_dark_triggers)) [INFO] piker.clearing._ems _ems.py:209 Dark order triggered for price 29824.5
Submitting order @ price 29825.0

May 09 13:53:07 (emsd, 34778, piker.clearing._ems.clear_dark_triggers)) [INFO] piker.clearing._ems _ems.py:257 removing pred for 2169fa61-b143-40cf-984a-3eb6d80bf213

May 09 13:53:07 (brokerd.kraken, 34735, piker.brokers.kraken.handle_order_requests)) [INFO] piker.brokers.kraken kraken.py:572 Received order request:
{'action': 'buy', 'oid': '2169fa61-b143-40cf-984a-3eb6d80bf213', 
'account': 'kraken.spot', 'time_ns': 1652118787952366919, 'reqid': None, 
'symbol': 'xbteur', 'price': 29825.0, 'size': 0.00419083}

May 09 13:53:08 (emsd, 34778, piker.clearing._ems.open_brokerd_trades_dialogue)) [INFO] piker.clearing._ems _ems.py:564 Received broker trade event:
{'account': 'kraken.spot',
 'broker_details': {'broker_time': '1652118788.621332', 'name': 'kraken'},
 'external': False,
 'filled': 0.00419083,
 'name': 'status',
 'reason': 'Order filled by kraken',
 'remaining': 0.0,
[...]

Issue #293, opened 2022-03-28
Add Support for Open Orders


iamzoltan:
It would be nice if we could add a message schema to deal with open orders stored with your broker, so that we could grab them and display them on piker.

In addition, I would like to add the possibility to save and reload dark orders and potentially paper engine orders as well.

iamzoltan:
@goodboy I think ill take a crack at this guy next, since I use kraken to hold open orders. It would be nice to be able to save dark orders too, something I can tackle next.

iamzoltan:
man, I am a stale ol' bish, well get on this asap

goodboy:
@iamzoltan we're basically primed for it after #349 lands if interested 😉

goodboy:
Prettty sure this was completely solved in https://github.com/pikers/piker/pull/374.

goodboy:
Follow up in https://github.com/pikers/piker/issues/384

PR #349, opened 2022-07-05
Kraken ws orders

goodboy:
WS based order management with fixes for #310, #296 and hopefully most of #290.

Leverage the new 3.10 `match:` syntax like crazy which turned out very nicely for making the msg loops way more grokable

Includes safeguard code to *cancel all persistent orders* found on startup until we get existing order loading working the EMS as per #293.

---
TODOs:
- [x] factor msg loop into lone async func: `handle_order_updates()`
- [x] get order "edits" fully working
- [x] handle all error cases
- [x] maybe break `{event: '[submit/edit/cancel]Orderstatus'}` blocks into a new sub-func?
- [x] test real live clearable orders #310 
- [x] test real live dark cleared orders #310
- [x] normalize symbols using assert pairs endpoint.
  - ~[ ] maybe we should delegate to `cryptofeed` for symbol normalization? bc kraken has a diff set for rest vs. ws apis.... (kill me)~ solved this for now using kraken's own asset info and pairs mapping tables.
- [x] use [`collections.ChainMap`](https://docs.python.org/3/library/collections.html#collections.ChainMap) to track the `dict`-oriented flow of order dialogs on the `kraken` api leg.

---
Follow up (mostly pertaining to the wishlist in #345):
- [x] what to do about balance transfers?
  - [x] eg. position is built out but then some asset is withdrawn, i guess the way would be to simply include withdrawals in `Position.size` logic?
  - [x] added basic withdrawal support in [c3ac1d4](https://github.com/pikers/piker/pull/349/commits/c3ac1d497fb627153d0b306530bdcd99fd74b529)
- ~[ ] using `piker.pp` immediate file updates, some we need to implement likely with the `trio` async file IO apis.~ added to #345 and will come in a follow up PR.
- [x] report the `userref` always being `0` in the `openOrders` ep to kraken here: https://support.kraken.com/hc/en-us/requests/new?ticket_form_id=96688


goodboy:
Order "edits" now fully working 🏄🏼

piker/brokers/kraken/broker.py
```diff
@@ -20,19 +20,19 @@
 '''
```
goodboy, line 20: Don't forget this file! 😂

goodboy:
Pretty sure everything is mostly working.

Just need to test some live clears and darks.

goodboy:
CI 💥

goodboy:
@iamzoltan @guilledk think this puppy is finally ready!

I've tested all live cases and we now have basic withdrawal compensation such that a position size will reflect such transactions in the accumulated size of any open position 🏄🏼

piker/brokers/kraken/README.rst
```diff
@@ -0,0 +1,64 @@
...
+your ``pps.toml`` file will have position entries like,
+
+.. code:: toml
+
+   [kraken.spot."xmreur.kraken"]
+   size = 4.80907954
+   ppu = 103.97000000
+   bsuid = "XXMRZEUR"
+   clears = [
+    { tid = "TFJBKK-SMBZS-VJ4UWS", cost = 0.8, price = 103.97, size = 4.80907954, dt = "2022-05-20T02:26:33.413397+00:00" },
```
goodboy, line 63: This schema actually changed in #365 so we might want to document the new column order and the `ppu` (aka breakeven price) stuff here as well.

PR #364, opened 2022-07-23
Add non-state-incremented calculation methods

goodboy:
Since we're going to need them anyway for desired features, add
2 new `Position` methods:
- `.calc_be_price()` which computes the breakeven cost basis price
  from the entries in the clears table.
- `.calc_size()` which just sums the clear sizes.

Add a `cost_scalar: float` control to the `.update_from_trans()` method
to allow manual adjustment of the cost weighting for the case where
a "non-symmetrical" model is wanted.

Go back to always trying to write the backing ledger files on exit, even
when there's an error (obvs without the `return` in the `finally:` block
f$#% up).

---
This is one more step toward tracking historical breakeven prices per position and thus eventually enabling pp "lifetime lines" in the chart UI as per #345.

### DIFF
--- a/piker/pp.py
+++ b/piker/pp.py
@@ -83,17 +83,18 @@ def open_trade_ledger(
         print(f'Ledger load took {time.time() - start}s')
         cpy = ledger.copy()
 
-    yield cpy
+    try:
+        yield cpy
+    finally:
+        if cpy != ledger:
+            # TODO: show diff output?
+            # https://stackoverflow.com/questions/12956957/print-diff-of-python-dictionaries
+            print(f'Updating ledger for {tradesfile}:\n')
+            ledger.update(cpy)
 
-    if cpy != ledger:
-        # TODO: show diff output?
-        # https://stackoverflow.com/questions/12956957/print-diff-of-python-dictionaries
-        print(f'Updating ledger for {tradesfile}:\n')
-        ledger.update(cpy)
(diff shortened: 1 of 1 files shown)


piker/pp.py
```diff
@@ -276,6 +277,62 @@ def lifo_update(
 
         return new_size, self.be_price
 
+    def calc_be_price(self) -> float:
```
goodboy, line 280: needs a doc string..

piker/pp.py
```diff
@@ -276,6 +277,62 @@ def lifo_update(
...
+            elif size_diff > 0:
+                # only an increaze in size of the position contributes
+                # the breakeven price, a decrease does not.
+
+                cost_basis += (
+                    # weighted price per unit of
+                    clear_price * abs(clear_size)
+                    +
+                    # transaction cost
+                    (copysign(1, new_size) * entry['cost'] * 2)
```
goodboy, line 311: the `* 2` bit should really be changed to allow user override as well as other modelling methods for exit cost estimation.

For an example `kraken` provides their fee tiers in the [asset info endpoint](https://docs.kraken.com/rest/#tag/Market-Data/operation/getAssetInfo), so it's definitely possible to dynamically calculate a breakeven level depending on the exit strategy.

guilledk (approved):

PR #365, opened 2022-07-26
Ppu history

goodboy:
Proper fixes and terminology to get us ready for historical position lifetimes as described in #345.

---
More to come in terms of explanation... or not 😂 

---
#### `pps.toml` format changes:
- [x] could we maybe just drop the top level `.size` and `be_price` values ?
  - so we can't (yet), because the ems uses the `Position` type internally as well and doesn't track all the
    clears (since that would be more msging overhead) so I think maybe having these is fine but we should probably
   change `.be_price` -> `.ppu`?
  - and instead just arrange the clears table to make these rolling values more obvious?, for eg:
```toml
[kraken.spot."xbteur.kraken"]
bsuid = "xbteur"
clears = [
 {  accum_size = 0.00639943, ppu = 19620.4996648139, cost = 0.32477, price = 19519.0, size = 0.00639943, dt = "2022-06-17T00:17:53.030664+00:00", tid = "TZVEXD-USARI-COY2ZR"},
```
which would be easiest to read if we sort clears reverse chronological so that the most recent trades (and thus their `ppu` and `accum_size` entries would be closest to the top?
- [x] it follows that maybe we should adjust the column order for each clears entry maybe to: `dt, ppu, accum_size, price, size, cost, tid`?
- ~[ ] i was hoping to have an indent per account (mentioned it in #345 as part of a custom encoder) such that it's distinguishable which pps are grouped together by backend, my only concern is then the clears table entries will be a bit cramped horizontally?~  **delaying this one again**

### DIFF
--- a/piker/brokers/ib/broker.py
+++ b/piker/brokers/ib/broker.py
@@ -383,7 +383,7 @@ async def update_and_audit_msgs(
                 symbol=ibppmsg.symbol,
                 currency=ibppmsg.currency,
                 size=p.size,
-                avg_price=p.be_price,
+                avg_price=p.ppu,
             )
             msgs.append(msg)
 
@@ -430,7 +430,7 @@ async def update_and_audit_msgs(
                 symbol=p.symbol.front_fqsn(),
                 # currency=ibppmsg.currency,
                 size=p.size,
-                avg_price=p.be_price,
+                avg_price=p.ppu,
             )
             if validate and p.size:
                 raise ValueError(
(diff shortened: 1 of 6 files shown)


guilledk (approved):
`"could we maybe just drop the top level .size and be_price values and instead just arrange the clears table to make these rolling values more obvious?"`
Sure, sounds good.
`"it follows that maybe we should adjust the column order for each clears entry maybe to: dt, ppu, accum_size, price, size, cost, tid?"`
Agreed.
`... "my only concern is then the clears table entries will be a bit cramped horizontally?"`
Mmmm not sure, id have to see an example.

Quick review of the code looks great.

PR #367, opened 2022-07-27
`ib`: live & paper accounts together, infra refinements

goodboy:
This is a first try at attempting to optimize for running a containerized paper gw alongside either another containerized live one or a local java app (tws or gw).

Currently it's just spawning two containers with `docker compose` and then hoping for the best in terms of being to use the vnc / data hacks since it seems both guis only show up in one of the two `x11vnc` instances 😂 

Also included is some logic to try and do symbol search queries in any running paper account if detected instead of the live account in an effort to not disrupt the data feeds (via extra load) for the live account where possible.


dockering/ib/jts_live.ini
```diff
@@ -0,0 +1,33 @@
...
+; RemoteHostOrderRouting=ndc1.ibllc.com
+; WriteDebug=true
+; RemotePortOrderRouting=4001
+; useRemoteSettings=false
+; tradingMode=p
+; Steps=8
+; colorPalletName=dark
+
+# window geo, this may be useful for sending `xdotool` commands?
+; MainWindow.Width=1986
```
goodboy, line 15: this is probably worth messing with in terms of more reliable VNC control stuff?

dockering/ib/run_x11_vnc_live.sh
```diff
@@ -0,0 +1 @@
+#!/bin/sh
```
goodboy, line 1: ah right, can probably drop this...

guilledk (approved):

PR #368, opened 2022-07-29
`kraken`: use `userref` field AND `reqid`, utilize `openOrders` sub for most msging

goodboy:
~Bleh, this is some mucking about~  A proper set of changes to use the `userref` field instead of `reqid` in submissions over the ws api.

Turns out you ~can't~ **can** use both! ~currently though I'm in the process of trying to show their API support why it's that a weird design that should be changed in order to make client implementations less complex...~

This moves fill handling, order edits, and cancellation event relaying to the `openOrders` sub block since after much discussion with support, this is more or less the *big boi sub* that indicates the most real-time state of their backend clearing engine.


guilledk (approved):
Havent tested but looks good, hehe also more pydantic removals eh?

goodboy:
> hehe also more pydantic removals eh?

🤫

PR #369, opened 2022-07-30
Drop `pydantic.create_model()` usage for `msgspec.defstruct()`

goodboy:
No idea how this got missed in #353 but likely had to do with not `pip uninstall`ing it 😂

### DIFF
--- a/piker/ui/_fsp.py
+++ b/piker/ui/_fsp.py
@@ -27,12 +27,13 @@ from itertools import cycle
 from typing import Optional, AsyncGenerator, Any
 
 import numpy as np
-from pydantic import create_model
+import msgspec
 import tractor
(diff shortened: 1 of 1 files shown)


guilledk (approved):

piker/ui/_fsp.py
```diff
@@ -153,12 +154,13 @@ async def open_fsp_sidepane(
     )
 
     # https://pydantic-docs.helpmanual.io/usage/models/#dynamic-model-creation
-    FspConfig = create_model(
-        'FspConfig',
-        name=name,
-        **params,
+    FspConfig = msgspec.defstruct(
```
goodboy, line 157: Heh, guess we should fix the comment link 😂

PR #390, opened 2022-08-25
Oneliner enable rpc modules on runtime open

### LINKED ISSUE #389: We are not enabling rpc modules on piker runtime start
While doing deribit backend work we found this line https://github.com/pikers/piker/blob/b5499b82251c76ff8c43461c77638d3cf7db7fe3/piker/_daemon.py#L220

We should be passing `_root_modules + enable_modules`, not just `_root_modules`.

guilledk:
Fixes #389


goodboy (approved):
nice

PR #391, opened 2022-08-25
Pull jsonrpc machinery out of deribit backend

guilledk:
We ended up doing all the `deribit.Client` methods work over a websocket transaprently from the caller persective, and machinery ended up real slick, so lets make it generic and pull it out of `deribit` backend to be used in other modules.

### DIFF
--- a/piker/data/_web_bs.py
+++ b/piker/data/_web_bs.py
@@ -179,8 +179,8 @@ async def open_jsonrpc_session(
         trio.open_nursery() as n,
         open_autorecon_ws(url) as ws
     ):
-        _rpc_id: Iterable = count(start_id)
-        _rpc_results: dict[int, dict] = {}
+        rpc_id: Iterable = count(start_id)
+        rpc_results: dict[int, dict] = {}
 
         async def json_rpc(method: str, params: dict) -> dict:
             '''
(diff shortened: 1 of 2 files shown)


piker/data/_web_bs.py
```diff
@@ -150,3 +153,85 @@ async def open_autorecon_ws(
...
+            _id = msg['id']
+
+            _rpc_results[_id] = {
+                'result': None,
+                'event': trio.Event()
+            }
+
+            await ws.send_msg(msg)
+
+            await _rpc_results[_id]['event'].wait()
```
goodboy, line 205: maybe flip all these to not have `_` since they're all local scope now.

piker/data/_web_bs.py
```diff
@@ -150,3 +153,85 @@ async def open_autorecon_ws(
...
+                    _rpc_results[msg.id] = {
+                        'result': None,
+                        'event': trio.Event()
+                    }
+
+                _rpc_results[msg.id]['result'] = msg
+                _rpc_results[msg.id]['event'].set()
+
+
+        n.start_soon(_recv_task)
```
goodboy, line 235: probably don't need it on `_recv_task()` either?

piker/data/_web_bs.py
```diff
@@ -150,3 +153,85 @@ async def open_autorecon_ws(
...
+
+        async def _recv_task():
+            '''
+            receives every ws message and stores it in its corresponding result
+            field, then sets the event to wakeup original sender tasks.
+            '''
+            async for msg in ws:
+                msg = dtype(**msg)
+
+                if msg.id not in _rpc_results:
```
goodboy, line 224: as discussed it's supposedly impossible for this block to enter but if it does, consider doing this instead :

```python
_rpc_results.setdefault(
    msg.id,
    {'result': None, 'event': trio.Event()}
)
```

and possibly consider adding a warning.
but i'd probably first start with all that code commented since we're never supposed to hit that case.

goodboy (approved):
couple style suggestions but looks great to me.

Issue #387, opened 2022-08-25
Backend test suite


goodboy:
In #331 comes a new backend test command `piker brokercheck` which i think we should run in CI against all backends that are officially supported.

Should be super simple to whip together with `pytest`.

Bonus points for extending things to be able to eventually spawn the tsdb and stream quotes from it use a data-layer set of tests 😉

goodboy:
Ended up moving the endpoint checks inside the new `.data.validate` from #489 and we have all the desired "service" tests already in CI but improved in that as well.

PR #393, opened 2022-08-27
Brokercheck an all backends

guilledk:
Begin work on #387

Do a `brokercheck` without logging on all backends, we pull backends from `piker.brokers.__brokers__`, I added `deribit` while I was at it cause I forgot to when doing the backend PR.

Currently not passing cause binance doesn't have a `trades_dialogue`, or `get_balance`.

Also should we log to console more like when running `brokercheck` on cli? This would require `log_cli` flag to be set on `pytest` config.

### DIFF
--- a/piker/brokers/__init__.py
+++ b/piker/brokers/__init__.py
@@ -30,6 +30,7 @@ __brokers__ = [
     'robinhood',
     'ib',
     'kraken',
+    'deribit'
 ]
 
 
--- a/piker/brokers/binance.py
+++ b/piker/brokers/binance.py
@@ -282,7 +282,7 @@ class Client:
 
 
 @acm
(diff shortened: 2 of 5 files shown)


goodboy:
> Currently not passing cause binance doesn't have a trades_dialogue, or get_balance.

Can we just maybe have a parameterization based on some kind of "supported" features log? I'm thinking we could check the backend subpackage module for an table that maybe denotes which features are supported and then we map those flags to a method set?

> Also should we log to console more like when running brokercheck on cli? This would require log_cli flag to be set on pytest config.

Don't see why not, you can set it in the test module or a `pytest.ini` iirc.

goodboy:
Replaced by warnings added in #489

PR #395, opened 2022-08-31
History view

goodboy:
~Very very~ pretty alpha right now but basics are working and we've got a bunch of follow up stuff in an issue (see below).

**NOTE** this branch must sync with the `tractor` branch in https://github.com/goodboy/tractor/pull/326

---
pic so i don't have to explain it..
screenshot-2022-08-30_17-56-04

**all follow up tasks moved to #400**

---
#### TODO (stuff i'm pretty sure we want):
- [x] basic linear region sync between history and rt chart
  - [x]  compute sampling ratio (eg. 1M to 1s OHLC is 60)
  - [ ] should a zoom out on the fast chart cause the slow chart to match once the fast range is `>=` the slow x-range
- [x] remove x-axis label from hist chart when mouse view
- [x] update the hist chart with rt data
  - [ ] how often to update, right now real-time, make it configurable?
  - [x] we don't want l1 labels right?
  - [x] show y-sticky for last price
- [x] pos nav in both the slow and fast chart
- [x] order/alert submission and edit from either chart
- [ ] key controls for hist view
  - [x] `r` should do the normal reset in hist view
  - [x] orders should work and be mirrored from rt view? (**yes on submit**)
  - [ ] do we want to *mirror* order staging on both charts in realtime? (**=> defer to <issue>**)
    - seems handy for checking bottoms against prior liquidity gaps?
  
----
### BUGz:
- [x] on boot, with `pikerd --tsdb` usage, the hist curve won't show right away until you pan it
- [x] no way to reset y-autorange on hist chart, so if you use y-axis scroll you're stuck in that mode
- [x] weird "black" pp icon when no pp on startup [see this comment](https://github.com/pikers/piker/pull/395#issuecomment-1240868287), and likely due to [4712163](https://github.com/pikers/piker/pull/395/commits/47121630c9a1d87843b70bbef5b3980635afd9ed) and [44a98f8](https://github.com/pikers/piker/pull/395/commits/44a98f809648d8e78376d7f3db64d27341a56866)

----
### Discussion / UI design:
- [x] region brush and side-line colors?
  - basically matches the side pane styling for now.
- [x] should the hist chart be "inside" the sidepane like the other charts?
  - [x] instead we should put the search pane adjacent to the slow chart and have it shown by default? **=> [313aa44](https://github.com/pikers/piker/pull/395/commits/313aa4475f290c28fbed4c2d54606bb649973279) and onwards**
- [ ] label with datums metrics?
  - eg. number of datums in view, amount of time that maps to
  - updated on interactions such as zoom, pan?
- [ ] right-click context menu to load hft data from a particular region of history if available from the backend
- [ ] modal level-triggered hotkey that when pressed makes the hist chart bigger?

### DIFF
--- a/piker/clearing/_allocate.py
+++ b/piker/clearing/_allocate.py
@@ -93,6 +93,9 @@ class Allocator(Struct):
         else:
             return self.units_limit
 
+    def limit_info(self) -> tuple[str, float]:
+        return self.size_unit, self.limit()
+
     def next_order_info(
         self,
 
--- a/piker/clearing/_ems.py
+++ b/piker/clearing/_ems.py
@@ -1109,7 +1109,7 @@ async def process_client_order_cmds(
                 # sometimes the real-time feed hasn't come up
                 # so just pull from the latest history.
                 if isnan(last):
-                    last = feed.shm.array[-1]['close']
+                    last = feed.rt_shm.array[-1]['close']
 
                 pred = mk_check(trigger_price, last, action)
 
--- a/piker/data/feed.py
+++ b/piker/data/feed.py
@@ -906,7 +906,7 @@ async def allocate_persistent_feed(
 
     # mem chan handed to broker backend so it can push real-time
     # quotes to this task for sampling and history storage (see below).
-    send, quote_stream = trio.open_memory_channel(10)
+    send, quote_stream = trio.open_memory_channel(616)
 
     # data sync signals for both history loading and market quotes
     some_data_ready = trio.Event()
--- a/piker/ui/_annotate.py
+++ b/piker/ui/_annotate.py
@@ -167,7 +167,6 @@ class LevelMarker(QGraphicsPathItem):
 
         '''
         level = self.get_level()
-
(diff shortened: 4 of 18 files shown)


guilledk (commented):
- I don't think history chart needs to be inside the sidepane.
- I'd like a label with datatum stats
- On the hotkey I don't have much opinion, I mostly use the mouse to move around.

goodboy:
oof also this...


screenshot-2022-09-08_22-13-38

ideally we don't have that tightness / overlap between the y-sticky and the pp line elements..

goodboy:
This also can't be right?

screenshot-2022-09-08_23-08-48

and i verified it against the fast cahrt; it appears the 1m bar isn't showing the same max/min ias the fast 1s in some cases?

guilledk (approved):
Tested for several days with no runtime errors! Awesome work man

goodboy:
@guilledk ok probably going to land this shortly and we'll handle any missing stuff posthumously 😂

PR #398, opened 2022-09-10
Oof, reverse clearing logic-routines in paper eng

goodboy:
Lul not sure how this exactly got broke but prolly was the final refactor in https://github.com/pikers/piker/pull/386..


guilledk (approved):

goodboy:
🏄🏼

Issue #400, opened 2022-09-19
History/slow chart view follow up


goodboy:
This is all the remnants and oustandings from #395 that we didn't clear up and/or discovered which need either fixes, further discussion

### Bugs:
- [ ] order markers on the slow chart seem to always stay in view despite the y-range when they should hide when out of view
- [ ] open and closing the search bar (now adjacent to the slow chart) causes the pp nav in the slow chart to have line, arrow and labels extended to where the y-axis was moved to until one updates the view with an interaction
  - this needs the same treatment as the `Nav` updates inside `GodWidget.load_symbol()`
  
### Enhancements / UX features
- [ ] should orders staged in one or the other chart be *real-time mirrored in the other*; iow as you move a line it moves proportionally in sync in the other chart?
  - see [draft commented code started inside `OrderMode.stage_order()`](https://github.com/pikers/piker/pull/395/files#diff-fa82dd622c1573f8025e48c8d3d1597ea2e9abaa6ee2a5357ff2bfab1a100081R298)
- [ ] should the linear range (aka *the god view* in the slow chart) set by the fast chart resize the view boundaries of the slow chart? As an example when you zoom out *so much* in the fast chart should the slow chart scale proportionally?
  - also see [example commented draft code in the callback def](https://github.com/pikers/piker/pull/395/files#diff-a91cd3e47717faa8e4317892b26b77102bec85a2f00109e0932cf289d5960d62R986)
- [ ] on the slow chart, should we place an inf line from the last y-sticky to make it easy to y-align the current price with history?
  
### UX style adjustments:
- [ ] should all slow chart order markers be set adjacent to the y-axis?
- [ ] should the slow chart always keep the linear region's y-axis side at least 2 cursor widths (or wtv) offset from that axis to make it easy to get non-linear-region scrolling interaction without zooming out?

PR #401, opened 2022-09-24
Ib 1m hist

goodboy:
Super WIP, but a start at doing 1m long term (slow) chart alongside our classic 1s OHLC in the fast chart.

---
TODO:
- [ ] actually a day (or more)'s worth of 1s day
- [ ] improving the the 1m loading algo to avoid slow waits on queries after 3 or 4
  - [ ] there's wip multi-task / don't wait for status even code already drafted
- [ ] init support for right-click-to-load-1s as is also mentioned in #400

---

maybe more to come?


piker/brokers/binance.py
```diff
@@ -396,6 +401,8 @@ async def get_ohlc(
             datetime,  # start
             datetime,  # end
         ]:
+            if timeframe != 60:
```
goodboy, line 404: So this is how we indicate that a `brokerd` can't deliver 1s OHLC.

piker/data/marketstore.py
```diff
@@ -642,30 +649,30 @@ async def tsdb_history_update(
...
-                for secs in (1, 60):
-                    ts = tsdb_arrays.get(secs)
-                    if ts is not None and len(ts):
-                        # these aren't currently used but can be referenced from
-                        # within the embedded ipython shell below.
-                        to_append = ohlcv[ohlcv['time'] > ts['Epoch'][-1]]
-                        to_prepend = ohlcv[ohlcv['time'] < ts['Epoch'][0]]
-
-            profiler('Finished db arrays diffs')
+            # for secs in (1, 60):
```
goodboy, line 668: Yeah just ignore most of this stuff since it's part of provisional tools we likely want for interactive tsdb mucking.

It relies on https://github.com/goodboy/tractor/pull/306 which is still far from ready 😂

piker/ui/_display.py
```diff
@@ -1069,6 +1117,12 @@ def update_region_from_pi(
 
             godwidget.resize_all()
 
+            await link_views_with_region(
```
goodboy, line 1120: this factoring was mostly for sanity and because i don't even really get what layer this whole subsystem fits in 😂

goodboy:
Just pushed a bad timeframe wiper hack @guilledk

guilledk (approved):

PR #402, opened 2022-09-30
Multi client order mgt

goodboy:
This originally was a set of bug fixes but turned into somewhat of an enhancement where we can now support multi-client management of an `emsd`'s orders.

---
### Summary:
- every client that connects to the ems for a given `fqsn` becomes a subscriber and thus un-hindered manager of orders. this fixes bugs to do with reconnecting clients who weren't previously receiving order updates as well as lays ground work for true multi-user trading on a given ems cluster.
- variety of face-palm-y `kraken` broker backend fixes
- teensie bit of logging adding to the paper engine

----
### TO TEST
start binance order mode with the paper engine and,
- [x] enter both live and dark orders and let some or none execute, disconnect the client, reconnect a new client, ensure that the old orders (both dark and live) show up and can be editted to get filled (without editing them to be clear-able immediately)
- [x] do the same but let some live orders clear without having any client connected (darks don't support this yet see below) and ensure on a chart-client reconnect the new pos is correct
- [x] for now you can probably verify that dark orders will not trigger when no client is connected
- [x] spawn more then one chart-client and be sure that changes from one are reflected in the other

---
### TODO:
- [ ] make the dark clearing loop also run as part of the lifetime of the `emsd` actor, offer this as a config option (**deferred to #405**)

### DIFF
--- a/piker/brokers/kraken/broker.py
+++ b/piker/brokers/kraken/broker.py
@@ -1015,11 +1015,10 @@ async def handle_order_updates(
                             fill_msg = BrokerdFill(
                                 time_ns=time.time_ns(),
                                 reqid=reqid,
-
-                                # action=action,  # just use size value
-                                # for now?
-                                size=vlm,
-                                price=price,
+                                # just use size value for now?
+                                # action=action,
+                                size=float(vlm),
+                                price=float(price),
 
                                 # TODO: maybe capture more msg data
                                 # i.e fees?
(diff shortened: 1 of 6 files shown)


guilledk:
Ok did the testing, test 1 and 2 work great.

3 as well dark dont get cleared but it seems to cancel the dark order it just deletes it.

4 opened a second chart and seems to have its own position tracking totally separate from the original chart.

goodboy:
> 4 opened a second chart and seems to have its own position tracking totally separate from the original chart.

but does it see the original chart's orders?

goodboy:
We need to pin to `tractor` master to make CI work again.

@guilledk cross-client-chart order mirroring is now workin 🏄🏼

guilledk (approved):
Tested and works great.

goodboy:
And yes this does resolve #402 🏄🏼

Issue #405, opened 2022-10-04
Run dark cleaing loop in `pikerd` subactor


goodboy:
As per the final unfinished item in #402, it's obviously ideal to be able to keep dark-cleared (aka algorithmic) order processing running when a client/UI is not connected to an `emsd`.

We need to persist both the rt quote feed and dark clearing task with any spawned `brokerd` order-dialog daemon-task.

Implementation should require a slight refactor to move components from `_emsd_main()` into `Router.maybe_open_brokerd_trades_dialogue()`.

---
#### Testing:
- [x] user should be able to submit a dark order, detach the chart/client and see the order clear on the `pikerd` console output
- [x] when a dark order is cleared when no chart is attached the order line should be removed from view on re-connect
- [x] we may want to add notification support as is found in the `OrderMode.on_exec()` handler to the daemon such that triggered orders can cause a notification that use can then use as reason (or source) for starting a client/chart?

goodboy:
I think this is done now right?

goodboy:
Yeah pretty sure this landed in https://github.com/pikers/piker/pull/408

I think it would be good however if we could write some tests for these mostly simple cases and just do some state checking on the paper engine to verify all the bullets above?

I think i'll just make a new issue to cover the test suite requirements (ping @algorandpa)

PR #406, opened 2022-10-05
Offline dark clearing

goodboy:
Resolves the requirements listed in #405 and requires the patch in https://github.com/goodboy/tractor/pull/329.

Makes the `emsd` run the dark clearing and relay loop tasks in inside the long lived `Router.nursery` and avoid more then a single task per instrument real-time feed except where needed by the paper clearing engine.

---
#### deats:
- `_emsd_main()` now on-demand spawns trade relay tasks per symbol and leaves them running even when no clients are connected to the EMS and now only directly determines the lifetime of the `process_client_order_cmds()` task
- includes some naive support for notifications when the `emsd` is running "headless" which is currently packed into `Router.client_broadcast()`
- solve the slight causality dilemma when a backend first boots:
  - when a client connects the first time it needs the current position and/or book state but on the first connect to some backend the book will be empty and needs to be populated with existing live orders, but the `translate_and_relay_brokerd_events()` needs to be first started so that existing open orders can be loaded and relayed to the client
  - **solution => make `Router.open_trade_relays()` deliver a `client_ready: trio.Event` which the client-caller-task sets once it opened a stream thus signalling the relay task to start and existing orders to be relayed to the now registered client**

---
#### test list:
see #405

### DIFF
--- a/piker/clearing/_ems.py
+++ b/piker/clearing/_ems.py
@@ -1275,13 +1275,48 @@ async def maybe_open_trade_relays(
 
 ) -> tuple:
 
-    relay, feed, client_ready = await _router.nursery.start(
-        _router.open_trade_relays,
-        fqsn,
-        exec_mode,
-        loglevel,
-    )
-    yield relay, feed, client_ready
+    def cache_on_fqsn_unless_paper(
+        router: Router,
+        fqsn: str,
+        exec_mode: str,  # ('paper', 'live')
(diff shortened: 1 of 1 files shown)


guilledk:
Tested on binance paper, works great.

goodboy:
@guilledk ready for review now as well.

added full tests check.

PR #409, opened 2022-10-19
Added support for JSONRPC requests coming from the server side

esmegl:
Looks like the `open_jsonrpc_session` code was only prepared to receive responses to requests that originated from the client, but not from the sever, which happens for example when subscribing to Deribit channels.

### DIFF
--- a/piker/brokers/deribit/api.py
+++ b/piker/brokers/deribit/api.py
@@ -95,8 +95,13 @@ class JSONRPCResult(Struct):
     error: Optional[dict] = None
     usIn: int 
     usOut: int 
-    usDiff: int 
+    usDiff: int
     testnet: bool
+
+class JSONRPCChannel(Struct):
+    jsonrpc: str = '2.0'
+    method: str
+    params: dict
 
 
 class KLinesResult(Struct):
--- a/piker/data/_web_bs.py
+++ b/piker/data/_web_bs.py
@@ -178,9 +178,6 @@ async def open_jsonrpc_session(
     request_type: Optional[type] = None,
     request_hook: Optional[Callable] = None,
     error_hook: Optional[Callable] = None,
-    timeout: int = 5,
-    timeout_hook: Optional[Callable] = None,
-    timeout_args: list = [],
 ) -> Callable[[str, dict], dict]:
 
     async with (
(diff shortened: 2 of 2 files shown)


piker/brokers/deribit/api.py
```diff
@@ -95,9 +95,14 @@ class JSONRPCResult(Struct):
     error: Optional[dict] = None
     usIn: int 
     usOut: int 
-    usDiff: int 
+    usDiff: int
     testnet: bool
 
+class JSONRPCChannel(Struct):
```
goodboy, line 101: am i missing where this new type is this being used?

if this is the msg that can also be received for non-request results but that show up on the ws channel maybe at least use this in the type annotation below?

See my comment below 😸

piker/data/_web_bs.py
```diff
@@ -172,7 +172,9 @@ class JSONRPCResult(Struct):
 async def open_jsonrpc_session(
     url: str,
     start_id: int = 0,
-    dtype: type = JSONRPCResult
+    response_type: type = JSONRPCResult,
```
goodboy, line 175: As per above comment could we maybe make this a union of this result and the new `JSONRPCChannel`?

So it would be:
```python
response_type: JSONRPCChannel | JSONRPCResult
```

further, if you wanted to get fancier / more modern with this, instead of doing the try / except in the msg loop below you could use a new `match msg:` syntax and then match against the msg type?

Check docs for this here:
https://peps.python.org/pep-0636/#adding-a-ui-matching-objects

obviously in the `JSONRPCChannel` case it would just be ignored / logged or wtv.

piker/data/_web_bs.py
```diff
@@ -219,18 +221,26 @@ async def recv_task():
...
+                    res = rpc_results.setdefault(
+                        msg.id,
+                        {'result': None, 'event': trio.Event()}
+                    )
+
+                    res['result'] = msg
+                    res['event'].set()
+
+                except TypeError:
+                    if request_type == None:
```
goodboy, line 239: `if request_type is None:` is more idiomatic i believe 😉

piker/data/_web_bs.py
```diff
@@ -219,18 +221,26 @@ async def recv_task():
             field, then sets the event to wakeup original sender tasks.
             '''
             async for msg in ws:
-                msg = dtype(**msg)
+                try:
```
goodboy, line 224: Yeah, so as mentioned above consider maybe trying out something like:

```python
match msg:
    case {
        'method': meth,
        'params': params,
     }:
         # ignore or log or wtv
```

you might be able to use your parsed object form as well but i think you might need a type hierarchy to make it work in `JSONRPCMsg(**msgdict)` style?

piker/data/_web_bs.py
```diff
@@ -219,28 +219,33 @@ async def recv_task():
...
+                        'result': _
+                    }:
+                        msg = response_type(**msg)
+
+                        if msg.id not in rpc_results:
+                            log.warning(f'Wasn\'t expecting ws msg: {json.dumps(msg, indent=4)}')
+
+                        res = rpc_results.setdefault(
+                            msg.id,
+                            {'result': None, 'event': trio.Event()}
```
goodboy, line 236: maybe simplify this to:

```python
res = rpc_results.setdefault(
    msg.id,
    {'result': msg, 'event': trio.Event()}
}
```

Also, i'm not entirely sure i follow the event setting here?
There either is, or isn't a waiting request right? In which case shouldn't you just assume the entry exists otherwise it's an error?

goodboy: Actually, just because I've gone through all the *tricks* with `match:` to make things more flat/succint, what y'all think about something more like this?

```python
match msg:
    case {
        'result': result,
        'id': mid,
    } if res_entry := rpc_results.get(mid):

        res_entry['result'] = response_type(**msg)
        res_entry['event'].set()

    case {
        'result': _,
        'id': mid,
    } if not rpc_results.get(mid):
        log.warning(f'Wasn\'t expecting ws msg: {json.dumps(msg, indent=4)}')

    case {
        'method': _,
        'params': _,
    } if request_hook:
        await request_hook(request_type(**msg))

    case _:
        log.warning(f'Unhandled JSON-RPC msg!?\n{msg}')
```

goodboy: 😿 dat coo peeps?

esmegl: Yep, done!

piker/data/_web_bs.py
```diff
@@ -42,117 +43,117 @@
...
+		return self
+
+	async def __anext__(self):
+		return await self.recv_msg()
 
 
 @asynccontextmanager
 async def open_autorecon_ws(
-    url: str,
+	url: str,
```
goodboy, line 139: what's with all the extra indent?

goodboy:
FYI rebasing to master should make CI green since it's fixed there now

piker/data/_web_bs.py
```diff
@@ -219,20 +225,49 @@ async def recv_task():
...
+                        }:
+                            log.warning(f'Recieved\n{error}')
+                            if error_hook:
+                                await error_hook(response_type(**msg))
+
+                        case _:
+                            log.warning(f'Unhandled JSON-RPC msg!?\n{msg}')
+
+            if cancel_scope.cancelled_caught:
+                await ws._connect()
```
goodboy, line 267: I think this whole recursively spawning tasks on timeout is a little confusing.

questions:
- if **this** task has timed out it shouldn't matter if the next try is scheduled async right?
- on timeout you run your `timeout_hook()` handler, does that need to be run async? if not, why not just do this all sync in an inf loop -> use a `while ws.connected()` (or wtv) and do a `break` once you get a result?

Further I think you could avoid passing a `timeout_args` and instead just make `timeout_hook` a `functools.partial` in the case you need to bind inputs.

goodboy (approved):
Looks good to me. 

Glad to see the `match:` stuff put to good use yet again 🏄🏼

guilledk (approved):

PR #412, opened 2022-10-28
Order ledger entries by processed datetime

goodboy:
To make it easier to manually read/decipher long ledger files this adds `dict` sorting based on record-type-specific (api vs. flex report) datetime processing prior to ledger file write.

- break up parsers into separate routines for flex and api record processing.
- add `parse_flex_dt()` for special handling of the weird semicolon stamps in flex reports.


goodboy:
Ahh nm just gonna land this in #401 since some changes in there rely on it and it's an internal `ib` ledger related patch.

PR #414, opened 2022-11-04
Agg feedz

goodboy:
Enhances our `piker.open_feed()` real-time quotes and history managment layer `piker.data.feed` to accept multi-fqsn inputs to deliver multi-symbol quote streams and a new internal data streaming abstraction/API: `.data.Flume` which provides the basis for real-time stream mangement, access and measure for the needs of real-time data-flow management and orchestration.

---
#### Synopsis
The final core-UX feature you always wanted as a chart trader is probably something like: 

*mult-instrument overlayed real-time and historical data feeds with simultaneous interaction and "current symbol" selectable order mode control*..

well, this **is finally within reach** 😎  and this patch add the "backend" work making it possible 🏄🏼 

---
#### Notes for manual testing
Ideally reviewers run the new feeds test set with `pytest tests/test_feeds.py`.
Note that you'll need to install the `piker_pin` branch of `tractor` in order for the test set to run green:
- the requirements file was updates in this patch: https://github.com/pikers/piker/pull/414/files#diff-4d7c51b1efe9043e44439a949dfd92e5827321b34082903477fd04876edb7552
- you can also simply switch to the remote `piker_pin` branch for `tractor` if installed in dev mode locally.

---
#### to land
- [x] fill out commit msg for 7abcb3e0 which was initial (half-working) patch to get basic funtionality 
- [x] port all consumer code in clearing, order mode, charting/graphics layer to expect this adjusted `Feed` api.
- [x] add basic per-`brokerd` multi-symbol real-time feeds working
  - enables `piker.open_feed(fqsns=['btcusdt.binance', 'ethusdt.binance']) as feed)` where the delivered `Feed` now has a `.flumes: dict[str, Flume]` which enables per-`fqsn` data flow access, mgmt, measure (see the historical [flume](https://en.wikipedia.org/wiki/Flume#Flow_measurement_flume) for idea behind this abstraction terminology)
  - [x] add a test in `test_feeds.py`
    - [x] `binance` multi-symbol case
    - [x] `kraken` multi-symbol case
    - [x] `kraken` currently seems to depend on a `brokers.toml` existing? we should fix this..

- [x] add cross-`brokerd` multi-feeds such that  `piker.open_feed(fqsns=['btcusdt.binance', 'xbtusdt.kraken']) will work with an aggregate receive channel delivering quotes from both backends?
    - [x] multi-brokerd` multi-sym case (`kraken` + `binance`)
    
---

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,6 +14,27 @@ on:
 
 jobs:
 
+  # test that we can generate a software distribution and install it
+  # thus avoid missing file issues after packaging.
+  sdist-linux:
+    name: 'sdist'
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Setup python
+        uses: actions/setup-python@v2
+        with:
+          python-version: '3.10'
+
+      - name: Build sdist
+        run: python setup.py sdist --formats=zip
+
+      - name: Install sdist from .zips
+        run: python -m pip install dist/*.zip
+
   testing:
     name: 'install + test-suite'
     runs-on: ubuntu-latest
--- a/dockering/ib/docker-compose.yml
+++ b/dockering/ib/docker-compose.yml
@@ -62,39 +62,39 @@ services:
       # - "127.0.0.1:4002:4002"
       # - "127.0.0.1:5900:5900"
 
-  ib_gw_live:
-    image: waytrade/ib-gateway:1012.2i
-    restart: always
-    network_mode: 'host'
+  # ib_gw_live:
+  #   image: waytrade/ib-gateway:1012.2i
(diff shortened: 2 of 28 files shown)


goodboy:
Le's try to land #421 before this.

goodboy:
### Test suite TODO: 
- [ ] we can't run `binance` tests from CI due to ip blacklisting on their end..
  - instead only run `kraken` tests in CI (using fixture logic for detection) and expect devs to run the full suite manually on PRs
- [ ] the initial fix for test failures on teardown was from the combo commits in https://github.com/goodboy/tractor/pull/347, but those fixes compeletely break `tractor` 😂, so we need either a better fix or to adjust the feeds tests to expect the streaming teardown gunk.

guilledk (approved):

PR #419, opened 2022-11-12
HOTFIX: Fix `_main()` arg back to `sym: str`

goodboy:
This slipped in early from #414 before merge and was likely due to cherry-picking from #417.

### DIFF
--- a/piker/ui/cli.py
+++ b/piker/ui/cli.py
@@ -174,7 +174,7 @@ def chart(
 
 
     _main(
-        syms=symbols,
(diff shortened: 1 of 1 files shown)


iamzoltan (approved):
quick hot fix works, but still having FLOWXBT issues, will sort out later

iamzoltan:
yea, this works for regular charting, but having issue with the tsdb (marketstore). will merge this and create an issue

iamzoltan:
also works with timeseries

PR #420, opened 2022-11-15
Multichartz

goodboy:
The UI follow up to #414.

**Hopefully this is the final core-UX feature you always wanted** 🏄🏼 

*mult-instrument overlayed real-time and historical data feeds with simultaneous interaction and "current symbol" selectable order mode control*

---
## THE CMD YOU NEED:
`piker chart btcusdt.binance ethusdt.binance xmrusdt.binance`

---
### First draft TODO
There are still tons of refinements  and bugs to fix as part of this first draft

#### Init basic funtionality / usage:
- [x] basic initial usage is to start a chart with multiple input symbols:
`piker chart btcusdt.binance ethusdt.binance xmrusdt.binance`
- [ ] dynamic managment of symbols per chart *view* (i.e. overlayed symbols can be added/removed from a running `LinkedSplits`
- [ ] more or less the idea of the **currently selected symbol** state as described in sub-sections below
  - [ ] select by y-axis
  - [ ] select by double click on curve?
  - [ ] select by hot key (some kinda `tab` or `space` global hotkey?)
- [ ] loading and saving of *watchlists* by name which contain multiple fqsns

#### Fast view
- [ ] selected feed should be closest to curve?
- [x] L1 labels can be simplified
  - [ ] place them all x-offset to the inner most y-axis?
  - [ ] only show (humanized) size on bid/ask not price?
- [ ] draw little line from each last label to curve?
- [ ] some kind of label system to easily know which curve is which feed?
  - [ ] last label should include fqsn?
  - [ ] each y-axis should have fqsn?
- [ ] for cross-brokerd symbols should we use a single axis?
  - [ ] how to determine this automatically? Do a *in the same range* calc?
- [ ] how should **last datum** graphics be offset, eg. should they be in reverse order of the y-axis from left-to-right?

#### History view
- [x] curves need to be epoch time aligned instead shm index aligned since obviously it will result in non-time aligned data sets as it is now using the `'index'` basis..
- [x]  fix `pg.LinearRegion` item to show all background overlay curves transparently

#### Graphics loop
- [ ] use single `Feed.open_multi_steam()` for the update task and have it look up appropriate graphics update apis per symbol-quote
- [ ] refactor a bunch of the charting/graphics apis to mostly to remove the global-ish `ChartPlotWidget` instance and API calls, instead moving all this state into `Flow` / `Flume` items and possibly add a `Viz` type to replace the `Flow` and instead make the `Flow` a table of associated `src data -> procesor` pipelines?
  - more to write up as individual subtasks here

#### Search UI/UX
- [ ] make search results and caching work with multi-symbol selections by the `GodWidget`
- [ ] all selected symbols should be shown in fqsn form in the window title bar
  - [ ] what about the symbol meta-data we used to have? -> maybe just put it in the feed label in the sidepane?
- [ ] allow switching between single-symbol selection 
- [ ] add support for multichart "feed groups" using the existing results tree view UI
  - [ ] adding and removing items from a overlay group
  - [ ] context menu options to add/remove symbols from a group
  - [ ] mouse UX (`shift-click-drag-release`?) to create a group from a single/flat results tree?
  - [ ] hot keys for adding a new search result to a group instead of a new widget set (maybe `ctl-o` or `ctl-enter` on result row?)
- [ ] add a "selected sub-symbol" UX where whatever highlighted symbol is in the results view changes the state of the "currently active contract":
  - [ ] highlight the feed's graphics curves and y-axis vs. the other overlays
  - [ ] activate order mode control for that symbol
  - [ ] show feed meta-data in sidepane label for selected
 
  
#### FSP (subcharts)
- [ ] should fsps like vlm which require a subchart be overlayed or only the currently selected shown by default?
  - we can make this configurable and maybe offer a sidepane drop down showing all combos and/or a toggle set of symbols from the overlay set?
- [ ] for ohlc main chart overlays how should we do coloring?

#### Order mode
- [ ] allow passing a `fqsns: list[str]` to `open_order_mode()` and having this setup the client loop and the ems tasks to use `feed.open_feed()` with the same `fqsns` passed through?
- [ ] pp line arrow-marker should sit on the specific y-axis?
- [ ] depending on "selected symbol" only show that pp?
- [ ] show active order summary markers (for those not in view) on each sub-yaxis
- [ ] support submitting orders per symbol **this is currently mega borked**
- [ ] how/when to show all pps?

### DIFF
--- a/piker/_profile.py
+++ b/piker/_profile.py
@@ -152,9 +152,14 @@ class Profiler(object):
             # don't do anything
             return cls._disabledProfiler
 
-        # create an actual profiling object
         cls._depth += 1
         obj = super(Profiler, cls).__new__(cls)
+        obj._msgs = []
+
+        # create an actual profiling object
+        if cls._depth < 1:
+            cls._msgs = []
+
         obj._name = msg or func_qualname
         obj._delayed = delayed
         obj._markCount = 0
@@ -174,8 +179,12 @@ class Profiler(object):
 
         self._markCount += 1
         newTime = perf_counter()
+        tot_ms = (newTime - self._firstTime) * 1000
         ms = (newTime - self._lastTime) * 1000
-        self._newMsg("  %s: %0.4f ms", msg, ms)
+        self._newMsg(
+            f"  {msg}: {ms:0.4f}, tot:{tot_ms:0.4f}"
+        )
+
         self._lastTime = newTime
 
     def mark(self, msg=None):
--- a/piker/brokers/ib/api.py
+++ b/piker/brokers/ib/api.py
@@ -161,9 +161,16 @@ _futes_venues = (
     'CME',
     'CMECRYPTO',
     'COMEX',
-    'CMDTY',  # special name case..
+    # 'CMDTY',  # special name case..
     'CBOT',  # (treasury) yield futures
 )
+
+_adhoc_cmdty_set = {
+    # metals
+    # https://misc.interactivebrokers.com/cstools/contract_info/v3.10/index.php?action=Conid%20Info&wlId=IB&conid=69067924
+    'xauusd.cmdty',  # london gold spot ^
+    'xagusd.cmdty',  # silver spot
+}
 
 _adhoc_futes_set = {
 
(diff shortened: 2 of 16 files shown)


goodboy:
This is going to get closed as it was replaced by a sequence of 7 (or so) smaller PRs ending with #455.

Also going to make a follow issue which will include the bullet set above rewritten and checked off as we finish follow up refinement tasks.

goodboy:
Replaced by the slew of PRs and follow up (copy pasted from here to start) task set in #461.

PR #424, opened 2022-11-23
Update notes_to_self.rst

sirdinkus:
adding IB protips section


goodboy:
add it to the wiki: https://github.com/pikers/piker/wiki

PR #426, opened 2022-12-01
Epoch time-align (indexing)

goodboy:
As I'm sure many have hear me complain, been working on epoch (time) aligning overlays to get #420 to where we'd likely want it.

Much more to come but putting this up for those who want to test.
Likely this will get merged with #420 eventually but breaking it out to have a testable and *working* version of multichartz so as to stay sane.. 🏄🏼


goodboy:
Replaced with a slew of other sub-PRs namely #451

Issue #427, opened 2022-12-01
Crash when an order is triggered when the notification daemon isn't running (linux only)


jaredgoldman:
When performing the following steps, Piker fails:
- Run pikerd with `sudo pikerd --tsdb --pdb`
- Load btc/binance charts with `piker -l cancel --tl -b binance chart --pdb btcusdt.binance`
- Try to place a limit order in paper mode
- Piker fails with the following error
```sh
Traceback (most recent call last):
  File "/home/algorandpa/dev/projects/piker/py-3.10/src/tractor/tractor/_root.py", line 218, in open_root_actor
    yield actor
  File "/home/algorandpa/dev/projects/piker/piker/_daemon.py", line 248, in maybe_open_runtime
    yield
  File "/home/algorandpa/dev/projects/piker/piker/_daemon.py", line 282, in maybe_open_pikerd
    yield portal
  File "/home/algorandpa/dev/projects/piker/piker/ui/_exec.py", line 181, in main
    await func(*((instance,) + args))
  File "/home/algorandpa/dev/projects/piker/piker/ui/_app.py", line 107, in _async_main
    async with (
  File "/home/algorandpa/dev/projects/piker/py-3.10/lib/python3.10/site-packages/trio/_core/_run.py", line 850, in __aexit__
    raise combined_error_from_nursery
  File "/home/algorandpa/dev/projects/piker/piker/ui/_display.py", line 1430, in display_symbol_data
    async with (
  File "/usr/lib/python3.10/contextlib.py", line 217, in __aexit__
    await self.gen.athrow(typ, value, traceback)
  File "/home/algorandpa/dev/projects/piker/piker/ui/order_mode.py", line 677, in open_order_mode
    async with (
  File "/home/algorandpa/dev/projects/piker/py-3.10/lib/python3.10/site-packages/trio/_core/_run.py", line 850, in __aexit__
    raise combined_error_from_nursery
  File "/home/algorandpa/dev/projects/piker/piker/ui/order_mode.py", line 919, in process_trades_and_update_ui
    await process_trade_msg(
  File "/home/algorandpa/dev/projects/piker/piker/ui/order_mode.py", line 1044, in process_trade_msg
    await notify_from_ems_status_msg(msg)
  File "/home/algorandpa/dev/projects/piker/piker/ui/_notify.py", line 86, in notify_from_ems_status_msg
    result = await trio.run_process(
[...]

PR #430, opened 2022-12-04
Catch notification daemon error

algorandpa:
closes issue #427

### DIFF
--- a/piker/ui/_notify.py
+++ b/piker/ui/_notify.py
@@ -82,6 +82,7 @@ async def notify_from_ems_status_msg(
             os.environ['DBUS_SESSION_BUS_ADDRESS'] = (
                 f'unix:path=/run/user/{_dbus_uid}/bus'
             )
+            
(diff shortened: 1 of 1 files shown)


piker/ui/_notify.py
```diff
@@ -93,6 +93,12 @@ async def notify_from_ems_status_msg(
...
             f"'{msg.pformat()}'",
-        ],
+        ],           
+        capture_stdout=True,
+        capture_stderr=True,
+        check=False,
     )
-    log.runtime(result)
+    if result.returncode != 0:
+        log.warn(f'No notification daemon installed stderr: {result.stderr}')
```
goodboy, line 102: my only concern might be line length on this, which i think right now python convention is 80 chars?

Not a huge deal tho and can always be tweaked later.

goodboy (approved):
Lit, nice easy fix 😎

guilledk (approved):

Issue #428, opened 2022-12-02
Daemon crashes if ~/.config/piker/marketshare folder isn't present


jaredgoldman:
When attempting to run the `pikerd --tsdb` on a fresh install, the daemon fails as the `~/.config/piker/marketstore` folder is not created automatically. We should add this on boot if the folder is not present. Fat stacks:

```sh
Traceback (most recent call last):
  File "/home/algorandpa/dev/repos/piker/py-3.10.8/bin/pikerd", line 33, in <module>
    sys.exit(load_entry_point('piker', 'console_scripts', 'pikerd')())
  File "/home/algorandpa/dev/repos/piker/py-3.10.8/lib/python3.10/site-packages/click/core.py", line 1130, in __call__
    return self.main(*args, **kwargs)
  File "/home/algorandpa/dev/repos/piker/py-3.10.8/lib/python3.10/site-packages/click/core.py", line 1055, in main
    rv = self.invoke(ctx)
  File "/home/algorandpa/dev/repos/piker/py-3.10.8/lib/python3.10/site-packages/click/core.py", line 1404, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/algorandpa/dev/repos/piker/py-3.10.8/lib/python3.10/site-packages/click/core.py", line 760, in invoke
    return __callback(*args, **kwargs)
  File "/home/algorandpa/dev/repos/piker/piker/cli/__init__.py", line 115, in pikerd
    trio.run(main)
  File "/home/algorandpa/dev/repos/piker/py-3.10.8/lib/python3.10/site-packages/trio/_core/_run.py", line 2010, in run
    raise runner.main_task_outcome.error
  File "/home/algorandpa/dev/repos/piker/piker/cli/__init__.py", line 100, in main
    ctn_ready, config, (cid, pid) = await n.start(
  File "/home/algorandpa/dev/repos/piker/py-3.10.8/lib/python3.10/site-packages/trio/_core/_run.py", line 1085, in start
    async with open_nursery() as old_nursery:
  File "/home/algorandpa/dev/repos/piker/py-3.10.8/lib/python3.10/site-packages/trio/_core/_run.py", line 850, in __aexit__
    raise combined_error_from_nursery
  File "/home/algorandpa/dev/repos/piker/piker/data/_ahab.py", line 379, in start_ahab
    async with portal.open_context(
[...]

PR #433, opened 2022-12-17
Add config dir on daemon startup

algorandpa:
Closes issue #428

### DIFF
--- a/piker/data/marketstore.py
+++ b/piker/data/marketstore.py
@@ -132,7 +132,10 @@ def start_marketstore(
 
     mktsdir = os.path.join(config._config_dir, 'marketstore')
 
-    # create when dne
+    # create dirs when dne
+    if not os.path.isdir(config._config_dir):
+        os.mkdir(config._config_dir)
+
     if not os.path.isdir(mktsdir):
         os.mkdir(mktsdir)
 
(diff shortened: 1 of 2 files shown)


goodboy (approved):
Looks right to me.

What do you think about a test for this?

I think more or less you'd just have to guarantee that the test runs before other tests that call `piker` CLI cmds and potentially create this dir?

piker/data/marketstore.py
```diff
@@ -132,7 +132,10 @@ def start_marketstore(
 
     mktsdir = os.path.join(config._config_dir, 'marketstore')
 
-    # create when dne
+    # create dirs when dne
```
goodboy, line 135: Wait, won't this fail since you're trying to make the `/marketstore` subdir *afterward*?

I'd luv to see a test verifying this creation sequence 😉

goodboy (changes requested):
Actually i *think* the order of creation might still result in the original error?

piker/data/marketstore.py
goodboy: Ah wait nm, estoy pendejo, i see the order is correct..

goodboy (approved):
Lol nm, it is correct. I misread the path `str` creation for the subdir as the actual dir creation code. 

This looks all good to me dawgy!

PR #438, opened 2023-01-09
Msgspec field ordering

guilledk:
Switch msgspec struct ordering to always have required fields first and optionals last


goodboy (approved):
Yup looks fine to me.

PR #446, opened 2023-01-29
First draft storage layer cli

goodboy:
Adds a `piker storage` subcmd with a `-d` flag to wipe a particular fqsn's time series (both 1s and 60s). Obviously this needs to be extended much more but provides a start point.

---
#### More to come:
Hopefully someone else might help pull this along 😉 

- [ ] add `piker storage sync` subcmd that maybe does gap analysis and backfilling?
- [ ] an alt what our current `piker storesh` to be instead `piker storage` which is passed without a sub-sub-cmd loads a REPL for interaction with the `Storage` APIs
- [ ] we probably should factor the `.data.marketstore.Storage` out into an api module and more generally add a `piker.storage` sub-pkg..

### DIFF
--- a/piker/data/cli.py
+++ b/piker/data/cli.py
@@ -19,7 +19,10 @@ marketstore cli.
 
 """
 from functools import partial
-from pprint import pformat
+from pprint import (
+    pformat,
+    pprint,
+)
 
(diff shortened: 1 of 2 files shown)


algorandpa (approved):
Thanks for this!

PR #447, opened 2023-01-30
Pregraphics formatters: `IncrementalFormatter`

goodboy:
In support of #420 (and the general effort towards having sanely sized PRs for others to review :joy:,) this introduces a "middleware" layer for our curve rendering graphics engine.

The main idea here is to formalize machinery for conducting *post `numpy.array`* (normally read from `ShmArray` buffers), *pre graphics-`QPainterPath`* formatting and incremental update, the main purpose of which is to speed up curve rendering cycles on real-time updated data flows.

---
The main abstractions to review include:
- the new `piker.ui._pathops.IncrementalFormatter` type hierarchy from ab1f15506 which introduces a the incremental update subsystem as mentioned above and which is primarily used from our `ui._flow.Renderer` layer: the subsys which does the work of transforming shm buffer data to the final `QPainterPath` graphics objects expected by `Qt`.
- cleaning of our curve graphics types (`BarItems` and `Curve`) to remove a bunch of old cruft-code and re-implement `.boundingRect()` methods to be sample step agnostic by relying on geometry calcs alone.

Further details about the (rather large) set of changes can be seen in commit msgs.

### DIFF
--- a/piker/ui/_flows.py
+++ b/piker/ui/_flows.py
@@ -565,26 +565,27 @@ class Flow(msgspec.Struct):  # , frozen=True):
 
         # XXX: SUPER UGGGHHH... without this we get stale cache
         # graphics that don't update until you downsampler again..
-        if reset:
-            with graphics.reset_cache():
-                # assign output paths to graphicis obj
-                graphics.path = r.path
-                graphics.fast_path = r.fast_path
+        # reset = False
+        # if reset:
+        #     with graphics.reset_cache():
+        #         # assign output paths to graphicis obj
+        #         graphics.path = r.path
+        #         graphics.fast_path = r.fast_path
 
-                # XXX: we don't need this right?
-                # graphics.draw_last_datum(
-                #     path,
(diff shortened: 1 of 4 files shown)


piker/ui/_flows.py
```diff
@@ -1208,18 +890,22 @@ def render(
         elif (
             append_length > 0
             and do_append
-            and not should_redraw
         ):
-            # print(f'{array_key} append len: {append_length}')
-            new_x = x_out[-append_length - 2:]  # slice_to_head]
-            new_y = y_out[-append_length - 2:]  # slice_to_head]
+            print(f'{array_key} append len: {append_length}')
```
goodboy, line 894: FYI this was a failed attempt at trying to get `QPainterPath` "appends" working by using graphics obj apis do do the appends at the path level instead of via a full redraw).

I'm leaving the commented code in for the future whenever we decide to take another try at it.

guilledk (approved):

PR #451, opened 2023-02-02
Epoch indexing and dataviz layer

goodboy:
Another in the push for #420.. and yes I realize this one is massive, but it's mostly internal data processing related stuff for improving curve overlays and making it easier to manage `Flume` data for visualization. 

This adds both *epoch time indexing*: meaning overlayed curves are now correctly aligned in an "epoch time stamp per sample" sense as well as a new `piker.ui._dataviz` layer which supports all the higher-level time series processing to enable sane update logic in `piker.ui._display` update tasks.

To focus, note the new modules:

`piker.ui._render.py` and `._dataviz.py` and `piker.data._pathops.py` which is a bunch of path processing related code moved out of the `ui.` subpkg.

Also we drop some reallly old code from the `piker.qt` subpkgs which was never really anything other then example snippets from ages ago.

---
As in #449 there are bunch of things yet to change you'll probably want **but which aren't yet proposed here**:
- still cluttered L1 labels
- order mode doesn't really work on the "front most" fqsn
- still terrible perf in terms of UX latency during chart interaction
- buncha debuggin print msgs on `piker chart` console that are cleaned up in later history


goodboy:
Huh I guess something musta changed in recent `msgspec`?

from ci:
```python
==================================== ERRORS ====================================
___________________ ERROR collecting tests/test_services.py ____________________
tests/test_services.py:20: in <module>
    from piker.clearing import (
/opt/hostedtoolcache/Python/3.10.9/x64/lib/python3.10/site-packages/piker/clearing/__init__.py:21: in <module>
    from ._client import open_ems
/opt/hostedtoolcache/Python/3.10.9/x64/lib/python3.10/site-packages/piker/clearing/_client.py:33: in <module>
    from ._messages import Order, Cancel
/opt/hostedtoolcache/Python/3.10.9/x64/lib/python3.10/site-packages/piker/clearing/_messages.py:232: in <module>
    class BrokerdStatus(Struct):
E   TypeError: Using a non-empty mutable collection ({'name': ''}) as a default value is unsafe. Instead configure a `default_factory` for this field.
```

guilledk (approved):

goodboy:
@guilledk once you get the protobuf dep fix in we can land this i think.

PR #455, opened 2023-02-03
Log linearized curve overlays

goodboy:
*The cross-symbol y-alignment you always wanted..*

screenshot-2023-02-03_08-17-43

Yah it's for #420 🏄🏼 now swapped with #461.

---
### wUt dIs?!
- overlays multiple time series using a "log-linearized returns transform", meaning all % returns (moves) on any symbol's chart are the same (from the reference point) as any other overlapped with it.
  - keeps support for multiple `Curve` per `PlotItem` y-range sorting
  - there are edge cases we handle:
    - any shorter series is aligned according to its *intersect* with the longer series it overlays with (in the case of > 2 this series is normally the one measured to have the *most dispersion*).
    - gaps are handled as though the pre-gap value is the value used as the ref-y-value
- reimplements all auto-y-ranging as part of a new `ChartView.interact_graphics_cycle()` which does the more efficient pipelining of `Viz` related graphics update calls per mouse event.
  - use this inside the `_display` loop as well.
- completely drops all the `Qt.Signal` stuff for interaction which results in more reliable, easier to understand, and generally faster interaction handling code 😂 

---
### Solved ToDo bugs:
- [x] we need to add back y-ranging on the vlm chart to make sure the RHS label is always in view (done in [0d45495](https://github.com/pikers/piker/pull/455/commits/0d45495a18aae12aaa3af3f546b78535a7c15bb2))
- [x] stupid gap on LHS of vlm fsp chart which still seems to be due to some kinda `PlotItem.layout` related setting?
  **LUL, fixed in [d4262c9](https://github.com/pikers/piker/pull/455/commits/d4262c975c37d736c97072d67a3d438c96ce6e44)**
screenshot-2023-02-03_10-48-22

- [x] get the immediate y-range incremental update on L1 working again?
  - not sure how this broke yet again, pretty sure it was just fine a a while back in 
  - **fixed and improved as of [fc0a793](https://github.com/pikers/piker/pull/455/commits/fc0a79331fd898b37e11ec1a980571f3a971d845)**

- [x] rewrite commit 325fe3cf which is duplicate of 325fe3cf from #453 after cherry picking it there.

----
### Final refinements
- [x] for a single fqsn using array-indexing the x-datetime labels for the cursor still aren't working.. (fixed by [269f659](https://github.com/pikers/piker/pull/455/commits/269f65988bfcbe50af501320296f6152cbc1c6e0))
- [x] bleh, still an issue with 3+ feeds on slow chart and overlays, eg.
screenshot-2023-02-10_14-40-34
  - HeH, figured it out -> we need to rescale any already scaled minors whenever the major (dispersion) needs to be rescaled (since any previously scaled minors are scaled to the previous major's range 🤦🏼 ) (**fixed in [1be5ae4](https://github.com/pikers/piker/pull/455/commits/1be5ae46a555793d8b2d0e26d8cf1c2e65d38804)**)
- [x] finish formalizing the `OverlayT` type and API and use lru caching on range (inverse) transform outputs
- [x] support the following 4 viewlist "modes" (finalized in 5e729dd)
  - [x] only one feed 
  - [x] ymxmn for each feed
  - [x] pin to dispersion major ([1be5ae4](https://github.com/pikers/piker/pull/455/commits/1be5ae46a555793d8b2d0e26d8cf1c2e65d38804))
  - [x] *pin from first* datum in view (rework in [16eda1e](https://github.com/pikers/piker/pull/455/commits/16eda1e02d839a6ef1f5a90dc72c20cfcf6b446b))

---
### FINAL FINAL bug fixes and tweaks:
- [x] fix upsampling (re-)triggering when using the `'r'` to reset chart hotkey in [1d649e5](https://github.com/pikers/piker/pull/455/commits/1d649e55caa107b5d1077eab2c481cfbaa530084)
- [ ] there's still an issue with de-synced 1m last bars?
  screenshot-2023-03-06_10-47-57

- [ ] lol, the alignment to "sigma major" default can definitely still cause a terrible config as per this example of bitcoin overlayed with a deribit call:
screenshot-2023-03-03_17-44-41
  - the issue is that the call option is clearly causing the bitcoin move to be mega squashed in the y-range 😂 
  - it might make sense in these cases to instead use a "leverage and delta adjusted" scalar for the opt contract so that you can see a "some what linear" translation of the deriv price vs. the underlying..
  - **yet another issue** seems to be that the "pin to dispersion major" curve idea doesn't work well when detecting intersects and the sigma major is the *shorter curve*?
    

https://user-images.githubusercontent.com/291685/223162334-6d881f36-446e-41f0-a6ea-389f9405ae32.mp4

### DIFF
--- a/piker/ui/_interaction.py
+++ b/piker/ui/_interaction.py
@@ -763,7 +763,7 @@ class ChartView(ViewBox):
 
         # NOTE: this value pairs (more or less) with L1 label text
         # height offset from from the bid/ask lines.
-        range_margin: float | None = 0.09,
+        range_margin: float | None = 0.06,
 
         bars_range: tuple[int, int, int, int] | None = None,
 
--- a/piker/ui/_interaction.py
+++ b/piker/ui/_interaction.py
@@ -956,6 +956,8 @@ class ChartView(ViewBox):
         debug_print: bool = False,
         do_overlay_scaling: bool = True,
         do_linked_charts: bool = True,
+
+        yranges: tuple[float, float] | None = None,
     ):
         profiler = Profiler(
             msg=f'ChartView.interact_graphics_cycle() for {self.name}',
@@ -1044,16 +1046,22 @@ class ChartView(ViewBox):
 
                 profiler(f'{viz.name}@{chart_name} `Viz.update_graphics()`')
 
-                out = viz.maxmin(i_read_range=i_read_range)
-                if out is None:
-                    log.warning(f'No yrange provided for {name}!?')
-                    return
-                (
-                    ixrng,
-                    read_slc,
-                    yrange
-                ) = out
-                profiler(f'{viz.name}@{chart_name} `Viz.maxmin()`')
+                yrange = yranges.get(viz) if yranges else None
+                if yrange is not None:
+                    # print(f'INPUT {viz.name} yrange: {yrange}')
+                    read_slc = slice(*i_read_range)
+
+                else:
+                    out = viz.maxmin(i_read_range=i_read_range)
+                    if out is None:
+                        log.warning(f'No yrange provided for {name}!?')
+                        return
+                    (
+                        _,  # ixrng,
+                        read_slc,
+                        yrange
+                    ) = out
+                    profiler(f'{viz.name}@{chart_name} `Viz.maxmin()`')
 
                 pi = viz.plot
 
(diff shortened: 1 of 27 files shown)


piker/ui/_interaction.py
```diff
@@ -340,6 +360,49 @@ async def handle_viewmode_mouse(
             view.order_mode.submit_order()
 
 
+class OverlayT(Struct):
```
goodboy, line 363: Note this code isn't used at all yet, it's just a draft for a more encapsulated API for the main logic inside `.interact_graphics_cycle()` below.

goodboy:
tweaked base branch to that from #471 for now just until we either drop those commits from this history or land that branch (which is slightly more critical 😂 )

goodboy:
Outstanding bullets added to #461.

Think this is ready to land 🏄🏼

PR #456, opened 2023-02-04
NixOS development envoirment

guilledk:
Add development shell environment file for devs on NixOS.

To load shell run:

`nix-shell develop.nix`

The shell will create a `virtualenv` and load it automaticaly, once inside proceed with the regular piker development install:

`pip install -e . -r requirements.txt`

In case you get a filesystem error about not enough space try increasing the `tmpfs` partition size, if using `systemd` can be done by `logind` config by adding the following to the system's `.nix` configuration:

```
  services.logind.extraConfig = ''
    RuntimeDirectorySize=50%
  '';
```

This will increase `tmpfs` partition size to 50% of ram, which should be alright for most cases (4GB of ram+).

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -78,7 +78,7 @@ loaded with::
 
     nix-shell develop.nix
 
-this will setup the required python envoirment to run piker, make sure to
+this will setup the required python environment to run piker, make sure to
 run::
 
     pip install -r requirements.txt -e .
--- a/README.rst
+++ b/README.rst
@@ -71,6 +71,19 @@ for a development install::
     source ./env/bin/activate
     pip install -r requirements.txt -e .
 
+install for nixos
+*****************
+for users of `NixOS` we offer a development shell envoirment that can be
+loaded with::
+
+    nix-shell develop.nix
+
+this will setup the required python envoirment to run piker, make sure to
+run::
+
+    pip install -r requirements.txt -e .
+
+once after loading the shell
 
 install for tinas
 *****************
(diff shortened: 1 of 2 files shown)


goodboy:
Ahh one of them brookly style *envoirment*s ehh?

README.rst
```diff
@@ -71,6 +71,19 @@ for a development install::
...
+    nix-shell develop.nix
+
+this will setup the required python environment to run piker, make sure to
+run::
+
+    pip install -r requirements.txt -e .
+
+once after loading the shell
 
 install for tinas
```
goodboy, line 88: @guilledk you mind also just dropping this tina section from the readme?

Maybe just toss it in a `tinaland/` subdir that we can focus on more later once we get more serious hackers on normie OSs?

develop.nix
```diff
@@ -0,0 +1,32 @@
...
+  ];
+  src = null;
+  shellHook = ''
+    # Allow the use of wheels.
+    SOURCE_DATE_EPOCH=$(date +%s)
+
+    # Augment the dynamic linker path
+    export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:${R}/lib/R/lib:${readline}/lib
+
+    export QT_QPA_PLATFORM_PLUGIN_PATH="${qt5.qtbase.bin}/lib/qt-${qt5.qtbase.version}/plugins";
```
goodboy, line 24: beauty, i imagine this was the mega-trick for `Qt` eh 😂

goodboy (approved):
Looks fantastic and simple.

Land when ready carnal!

develop.nix
iamzoltan: slick

PR #457, opened 2023-02-11
Use new msgspec default factories

guilledk:
Added on `msgspec` v0.13.0, for more information about this change see: https://github.com/jcrist/msgspec/pull/274


goodboy (approved):
Huh, interesting they went with the `@dataclass` style. I gotta read that changelog in more detail.

PR #460, opened 2023-02-13
Fix crash on notification daemon not found

guilledk:
Fix case where user has no notification daemon installed and an order clears, we crash due to `notify-send` command not present, previously @algorandpa tried to address this issue but only caught the case where the `notify-send` command exits with non zero exit code, now I add a try-catch around the offending code and log a warning on error, avoiding crash.


goodboy (approved):
Nice catch 🏄🏼

PR #464, opened 2023-02-19
Elasticsearch integration

esmegl:
- Added elasticsearch configurations on dockering folder
- Added `--es` command on `piker/cli/__init__.py`
- Created elasticsearch script to start elasticsearch instance
- Modified ahab to support elasticsearch logs and init procedure
- Created new tests for elasticsearch and markestore initialization


piker/cli/__init__.py
```diff
@@ -108,6 +115,27 @@ async def main():
                     f'config: {pformat(config)}'
                 )
 
+            if es:
```
guilledk, line 118: Looking good, but we need to move the database initialization code inside the `open_pikerd` context manager, to make it more generic and not duplicate code on tests.

esmegl: Ok, working on it

piker/data/_ahab.py
```diff
@@ -143,27 +143,37 @@ async def process_logs_until(
...
                 try:
-                    record = json.loads(entry.strip())
+                    record = json.loads(entry)
+
+                    if 'msg' in record:
+                        msg = record['msg']
+                    elif 'message' in record:
+                        msg = record['message']
+                    else:
+                        raise KeyError('Unexpected log format')
```
goodboy, line 155: dump the `msg` contents here in the exception message as well so that if/when this breaks we have a clue what the problematic msg was.

piker/data/_ahab.py
```diff
@@ -143,27 +143,37 @@ async def process_logs_until(
...
+
+                    level = record['level']
+
                 except json.JSONDecodeError:
-                    if 'Error' in entry:
-                        raise RuntimeError(entry)
-                    raise
+                    # if 'Error' in entry:
```
goodboy, line 160: Yeah this was obviously oriented around marketstore, so if it's not generic enough we should just drop it.

piker/data/_ahab.py
```diff
@@ -143,27 +143,37 @@ async def process_logs_until(
...
-                    getattr(log, level, log.error)(f'{msg}')
+                    getattr(log, level.lower(), log.error)(f'{msg}')
 
-                    # print(f'level: {level}')
-                    if level in ('error', 'fatal'):
+                    if level == 'fatal':
                         raise ApplicationLogError(msg)
 
-                if patt in msg:
+                if patt_matcher(msg):
```
goodboy, line 176: I like this predicate usage but we should maybe put a comment near the type sig at the top of this routine mentioning that this is a predicate func for matching log msgs emitted by the underlying containerized app yah?

piker/data/_ahab.py
```diff
@@ -143,27 +143,37 @@ async def process_logs_until(
...
                     seen_so_far.add(entry)
                     if bp_on_msg:
                         await tractor.breakpoint()
 
-                    getattr(log, level, log.error)(f'{msg}')
+                    getattr(log, level.lower(), log.error)(f'{msg}')
 
-                    # print(f'level: {level}')
-                    if level in ('error', 'fatal'):
+                    if level == 'fatal':
```
goodboy, line 173: my only concern is if this is gonna cause benign errors triggering crashes with `marketstore`?

@guilledk you tested this with the `--tsdb` stuff?

piker/data/elasticsearch.py
```diff
@@ -0,0 +1,108 @@
...
+from trio_websocket import open_websocket_url
+import pendulum
+import purerpc
+
+if TYPE_CHECKING:
+    import docker
+    from ._ahab import DockerContainer
+
+from .feed import maybe_open_feed
+from ..log import get_logger, get_console_log
```
goodboy, line 53: trying to stick with a multi-line tuple style for any more then one import for a module if that's cool, like above for the `typing` imports.

piker/data/elasticsearch.py
```diff
@@ -0,0 +1,108 @@
...
+        network='host',
+        detach=True,
+        remove=True,
+    )
+
+    return (
+        dcntr,
+        {},
+        # expected startup and stop msgs
+        lambda msg: msg == "started",
```
goodboy, line 106: maybe make the input names here more explicit so that it's clear which lambda is which to a reader of this code:

eg. `start_msg` and `stop_msg`

piker/data/marketstore.py
```diff
@@ -194,8 +194,8 @@ def start_marketstore(
         _config,
 
         # expected startup and stop msgs
-        "launching tcp listener for all services...",
-        "exiting...",
+        lambda msg: "launching tcp listener for all services..." in msg,
```
goodboy, line 197: same thing here as mentioned in previous comment.

goodboy (approved):
Other then some slight nitpicks this looks good to me.

I'll let @guilledk guide the patch set to completion / merge.

I'd like to also give this at least a serious manual test (once you flip it to a non-draft PR) before landing, though if y'all feel comfortable adding a small integration test (maybe to `test_services.py`?) then that would be even better 🏄🏼

piker/data/_ahab.py
guilledk: I did a manual test starting with both `--es` and `--tsdb` and both opened ok, didnt do more extensive testing than that tho.

guilledk: Tests for initialization of `elastic` & `marketstore` are passing.

guilledk (approved):

tests/test_databases.py
```diff
@@ -0,0 +1,64 @@
...
+
+    '''
+    log = get_logger(__name__)
+
+    async def main():
+        # port = 5995
+
+        async with (
+            open_test_pikerd(
+                loglevel='info',
```
goodboy, line 27: you can pass this in from the `loglevel` fixture now which was added in #465

tests/test_databases.py
```diff
@@ -0,0 +1,64 @@
...
+
+    '''
+
+    log = get_logger(__name__)
+
+    async def main():
+        port = 19200
+
+        async with open_test_pikerd(
+            loglevel='info',
```
goodboy, line 56: same here!

tests/test_databases.py
```diff
@@ -0,0 +1,64 @@
...
+
+from typing import AsyncContextManager
+
+from piker._daemon import Services
+from piker.log import get_logger
+
+from elasticsearch import Elasticsearch
+from piker.data import marketstore
+
+def test_marketstore(
```
goodboy, line 12: maybe change the names of these tests to be something like `test_<dbname>_startup_and_version()` ? Just so we can distinguish them from other more (presumably) specific tests in the future?

piker/data/elastic.py
```diff
@@ -0,0 +1,120 @@
...
+- websocket client for subscribing to write triggers
+- todo: tick sequence stream-cloning for testing
+
+'''
+from __future__ import annotations
+from contextlib import asynccontextmanager as acm
+from pprint import pformat
+from typing import (
+    Any,
+    # Optional,
```
goodboy, line 31: you can probably just drop these commented types ( i realize it's from the copy paste 😂 )

piker/data/elastic.py
```diff
@@ -0,0 +1,120 @@
...
+        return health[0]['status'] == 'green'
+
+    async def stop_matcher(msg: str):
+        return msg == 'closed'
+
+    return (
+        dcntr,
+        {},
+        # expected startup and stop msgs
+        start_matcher,
```
goodboy, line 118: Hmm these are really more like generic *predicate* functions now instead of just being msg matchers. I wonder the naming will confuse anyone going foward? Maybe it's not imporant for now 😂

piker/data/elastic.py
```diff
@@ -0,0 +1,120 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+'''
+``elasticsearch`` integration.
+
+- client management routines
```
goodboy, line 20: Is this from the copy paste as well?

If so maybe just wipe these bullets.

piker/data/_ahab.py
```diff
@@ -135,39 +137,52 @@ async def process_logs_until(
         seen_so_far = self.seen_so_far
 
         while True:
-            logs = self.cntr.logs()
+            try:
+                logs = self.cntr.logs()
+            except docker.errors.NotFound:
```
goodboy, line 142: you can compact these 2 except blocks into one using tuple syntax i believe?

```python
except (
    docker.errors.NotFound,
    docker.errors.APIError,
):
    return False
```

goodboy (changes requested):
@esmegl lookin super duper good!

I have a couple nitpicks mostly naming and style related and then I think this is definitely ready to land 😎

guilledk (approved):

piker/data/_ahab.py
```diff
@@ -124,7 +124,9 @@ def __init__(
 
     async def process_logs_until(
         self,
-        patt: str,
+        # this is a predicate func for matching log msgs emitted by the
+        # underlying containerized app
+        patt_matcher: Callable[[str], bool],
```
goodboy, line 129: I think the typing here is techincally an `Awaitable` now, but it's not a big deal and i don't want this to delay the patch.

piker/data/elastic.py
```diff
@@ -0,0 +1,109 @@
...
+            )).json()
+
+        except OSError:
+            log.error('couldnt reach elastic container')
+            return False
+
+        log.info(health)
+        return health[0]['status'] == 'green'
+
+    async def stop_matcher(msg: str):
```
goodboy, line 100: returns a `bool`. again though, not a huge deal and shouldn't block this.

tests/test_databases.py
```diff
@@ -0,0 +1,66 @@
...
+from piker.log import get_logger
+
+from elasticsearch import Elasticsearch
+from piker.data import marketstore
+
+def test_marketstore_startup_and_version(
+    open_test_pikerd: AsyncContextManager,
+    loglevel,
+):
+
```
goodboy, line 16: don't need this newline

goodboy (approved):
A couple more minor suggestions but nothing that should necessarily block this if y'all are happy with it 👍🏼 

@guilledk @esmegl great work yet again and nice to see a new integration!

PR #465, opened 2023-02-21
`loglevel` to `open_test_pikerd()` via `--ll <level>` flag

goodboy:
Like it sounds, this jigs our test harness to include:
- a new `loglevel: str` session scoped fixture for downstream use, pulled directly from the `--ll <level: str>` input by the user via CLI
- bound in `loglevel` value to `open_test_pikerd()` async mngr to allow controlling full service tree logging from `pytest` CLI

Also included are some minor tweaks to remove some warnings from the summary, mostly to do with `tractor` API deprecations, and some moving of old `conftest` fixture code to the `test_questrade` module.

### DIFF
--- a/piker/clearing/_client.py
+++ b/piker/clearing/_client.py
@@ -175,6 +175,7 @@ async def relay_order_cmds_from_sync_code(
 async def open_ems(
     fqsn: str,
     mode: str = 'live',
+    loglevel: str = 'error',
 
 ) -> tuple[
     OrderBook,
@@ -244,6 +245,7 @@ async def open_ems(
                 _emsd_main,
                 fqsn=fqsn,
                 exec_mode=mode,
+                loglevel=loglevel,
 
             ) as (
                 ctx,
--- a/piker/brokers/__init__.py
+++ b/piker/brokers/__init__.py
@@ -20,10 +20,6 @@ Broker clients, daemons and general back end machinery.
 from importlib import import_module
 from types import ModuleType
 
-# TODO: move to urllib3/requests once supported
-import asks
-asks.init('trio')
-
 __brokers__ = [
     'binance',
     'ib',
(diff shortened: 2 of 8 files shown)


algorandpa (approved):

guilledk (approved):

jaredgoldman (changes requested):
@goodboy You should add log level to `test_runtime_boot` as it's instantiation is logging at `info` by default. Also there is a bunch of market data being dumped in the console:

goodboy:
> `test_runtime_boot`

The log dumping to console is on purpose actually since we might want to assert a schema from those msgs eventually using `pytest`'s streams capture.

I will change the test for sure 👍🏼

goodboy:
> You should add log level to test_runtime_boot as it's instantiation is logging at info by default. Also there is a bunch of market data being dumped in the console:

As per chat discussion, the issue isn't this test func since the fixture takes care of binding in the log level. So it's got to be an issue elsewhere (probably hardcoded info level?).

I'll check.

goodboy:
Yup, had to patch the `clearing._client.open_ems()` to passthrough a `loglevel: str` and then adjust the service tests to passthrough the fixture value.

All there now please review!

piker/clearing/_client.py
```diff
@@ -175,6 +175,7 @@ async def relay_order_cmds_from_sync_code(
 async def open_ems(
     fqsn: str,
     mode: str = 'live',
+    loglevel: str = 'error',
```
goodboy, line 178: Yeah i think ideally in the longer run we have a better "tagging" log level system such that we can flick on particular "layers" of the stack with different levels of "granularity" (which is more like the built-in python leveling system).

I think some of this is discussed in https://github.com/goodboy/tractor/issues/244 and

jaredgoldman:
Looking good now @goodboy

jaredgoldman (approved):

PR #468, opened 2023-02-25
Update nix dev env to 3.11.1

guilledk:
Can't really merge this yet cause `numba` hasnt updated its max version requirement to allow `3.11.1`, and nix package `python311` is hardcoded to `3.11.1`

:clown_face:

### DIFF
--- a/develop.nix
+++ b/develop.nix
@@ -1,5 +1,5 @@
 with (import <nixpkgs> {});
-with python310Packages;
+with python311Packages;
 stdenv.mkDerivation {
   name = "pip-env";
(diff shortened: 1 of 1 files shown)


goodboy:
@guilledk i'm on nix with unstable channels now so i'm going to try making this work again 🏄🏼

goodboy:
Unfortunately this `venv` hack approach i don't think is going to work once we're in a state like pkging is atm what with `numba` requiring `numpy<=1.24`..

I'm going to close this favor of the more correct (and flakes based) approach coming in #533

PR #469, opened 2023-02-26
Fix `loglevel` not getting propagated to `emsd`

guilledk:
Quick one, while debugging decimal issues we realized no logs were getting printed from `emsd`, easy fix.


goodboy (approved):
yupppp!

not sure how this got missed 😂

PR #462, opened 2023-02-15
 Paper trade improvements

algorandpa:
- Added ledger and positioning support for paper mode
- Ensure we immediately write to ledger and update position in real time
- Added local txn state in Paperboi

---
#### Test suite todo:
- [x] clear a paper order, ensure the pp shows up in `pps.toml`, stop and start `pikerd`, ensure the pp is relayed on `open_ems()` connect
- [ ] ensure after a return to a net zero position, the pps.toml does not get a leaked account entry as per comment https://github.com/pikers/piker/pull/462#issuecomment-1438852266
- [x] when loading a paper position after a full restart of the service stack (i.e. shutting down and restarting pikerd) if you make a new clear to begin "closing out" (i.e. reducing the dst asset) the paper engine is not sending the correct `BrokerdPosition` state as per the comment https://github.com/pikers/piker/pull/462#discussion_r1113400683, as such we need to add test code to process the `trade_stream: tractor.MsgStream` msgs and ensure the eventual last position msg matches the state in the `pps.toml`

---
#### Future dark-order and alert storage todo test suite (maybe to do in a follow up PR):
~submit an order that does not clear, stop and start `pikerd`, ensure order is re-registered in `emsd`~ => **deferred to  #463**
  ~will likely requires  [`Order`](https://github.com/pikers/piker/blob/4833d56ecb5ee33076cd245ad129659a102423f4/piker/clearing/_messages.py#L67) serialization to file system and load~  => **deferred to #464**

PR #470, opened 2023-02-27
Fixed float dust bug on zero position

guilledk:
In #462 we discovered a bug on the positioning system when trying to get to a zero position state, after investigation we found out it was caused by float precision, the solution is to truncate based on the maximun supported on a symbol by symbol basis.

### DIFF
--- a/piker/brokers/kraken/broker.py
+++ b/piker/brokers/kraken/broker.py
@@ -469,8 +469,7 @@ async def trades_dialogue(
         with (
             open_pps(
                 'kraken',
-                acctid,
-                write_on_exit=True,
+                acctid
             ) as table,
 
             open_trade_ledger(
--- a/piker/brokers/kraken/broker.py
+++ b/piker/brokers/kraken/broker.py
@@ -48,6 +48,7 @@ from piker.pp import (
     open_trade_ledger,
     open_pps,
 )
+from piker.data._source import Symbol
 from piker.clearing._messages import (
     Order,
     Status,
@@ -1196,10 +1197,21 @@ def norm_trade_records(
         }[record['type']]
 
         # we normalize to kraken's `altname` always..
-        bsuid = norm_sym = Client.normalize_symbol(record['pair'])
+        bsuid, pair_info = Client.normalize_symbol(record['pair'])
+        fqsn = f'{bsuid}.kraken'
+
+        mktpair = Symbol.from_fqsn(
+            fqsn,
+            info={
+                'lot_size_digits': pair_info.lot_decimals,
+                'tick_size_digits': pair_info.pair_decimals,
+                'asset_type': 'crypto',
+            },
+        )
 
         records[tid] = Transaction(
-            fqsn=f'{norm_sym}.kraken',
+            fqsn=fqsn,
+            sym=mktpair,
             tid=tid,
             size=size,
             price=float(record['price']),
--- a/piker/brokers/kraken/broker.py
+++ b/piker/brokers/kraken/broker.py
@@ -474,7 +474,7 @@ async def trades_dialogue(
 
             open_trade_ledger(
                 'kraken',
-                acctid,
+                acctid
             ) as ledger_dict,
         ):
             # transaction-ify the ledger entries
--- a/piker/clearing/_paper_engine.py
+++ b/piker/clearing/_paper_engine.py
@@ -38,6 +38,7 @@ import tractor
 
 from .. import data
 from ..data.types import Struct
+from ..data._source import Symbol
 from ..pp import (
     Position,
     Transaction,
@@ -81,6 +82,7 @@ class PaperBoi(Struct):
     _reqids: bidict
     _positions: dict[str, Position]
     _trade_ledger: dict[str, Any]
+    _syms: dict[str, Symbol] = {}
 
     # init edge case L1 spread
     last_ask: tuple[float, float] = (float('inf'), 0)  # price, size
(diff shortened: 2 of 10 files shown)


piker/clearing/_paper_engine.py
```diff
@@ -530,7 +535,7 @@ async def trades_dialogue(
 
     ):
 
-        with open_pps(broker, 'paper', False) as table:
+        with open_pps(broker, 'paper') as table:
```
goodboy, line 538: not quite following why you're not writing immediately here any more?

piker/data/_source.py
```diff
@@ -156,14 +156,14 @@ def from_broker_info(
     ) -> Symbol:
 
         tick_size = info.get('price_tick_size', 0.01)
-        lot_tick_size = info.get('lot_tick_size', 0.0)
+        lot_size = info.get('lot_tick_size', 0.0)
```
goodboy, line 159: much better name 🥳

piker/pp.py
```diff
@@ -680,10 +693,10 @@ def write_config(self) -> None:
         '''
         # TODO: show diff output?
         # https://stackoverflow.com/questions/12956957/print-diff-of-python-dictionaries
-        print(f'Updating ``pps.toml`` for {path}:\n')
-
+        log.info(f'Updating ``pps.toml`` for {path}:\n')
         # active, closed_pp_objs = table.dump_active()
         pp_entries = self.to_toml()
+        log.info(pp_entries)
```
goodboy, line 699: maybe add a header to this msg?

`log.info(f'Current positions:\n{pp_entries}')`

piker/pp.py
```diff
@@ -618,7 +631,7 @@ def dump_active(
             # XXX: debug hook for size mismatches
             # qqqbsuid = 320227571
             # if bsuid == qqqbsuid:
-            #     breakpoint()
+            #     xbreakpoint()
```
goodboy, line 634: drop this?

piker/pp.py
```diff
@@ -189,9 +191,20 @@ def to_pretoml(self) -> tuple[str, dict]:
...
         # it via the trades ledger..
         # drop symbol obj in serialized form
-        s = d.pop('symbol')
+        s = d.get('symbol')
         fqsn = s.front_fqsn()
 
+        broker, key, suffix = unpack_fqsn(fqsn)
+        sym_info = s.broker_info[broker]
+
+        d['symbol'] = {
```
goodboy, line 200: excelente nieri

piker/data/_source.py
```diff
@@ -254,6 +254,12 @@ def iterfqsns(self) -> list[str]:
 
         return keys
 
+    def decimal_quant(self, d: Decimal):
+        digits = self.lot_size_digits
+        return d.quantize(
```
goodboy, line 259: 💥

goodboy (approved):
This looks excellent yet again @guilledk.

I have a few comments but nothing blocking afaict.

Great work and feel free to merge into the target dev branch whenever ready.
I presume you might need to rebase after the suggested changes made in the last review on #462 land.

piker/clearing/_paper_engine.py
guilledk: I'm pretty sure that `open_pps` call is just meant to load the `pps.toml` file into the `_positions` module var and then iterate on it to send position messages to the ems.

piker/brokers/kraken/broker.py
```diff
@@ -470,6 +470,7 @@ async def trades_dialogue(
             open_pps(
                 'kraken',
                 acctid,
+                write_on_exit=False
```
jaredgoldman, line 473: We don't need this anymore :)

jaredgoldman (approved):
LGTM other than the first comment

goodboy:
I'm just gonna manually test with the same scenario I got in #462, presuming it's resolved I'll merge 🏄🏼

goodboy:
@guilledk another thing I think should change is to just put the `.symbol` section fields under the position entry no?

screenshot-2023-03-01_18-13-13

when are we ever going to need to have a separate section?
- a given `asset_type = "crypto" price_tick_size = 0.01 lot_tick_size = 1e-5` set is always going to be specific to the particular market/symbol/fqsn for a given broker right?
- i don't see any case where this extra sibling section is going to be shared/re-used right?

piker/pp.py
```diff
@@ -957,7 +981,8 @@ def open_pps(
             expiry = pendulum.parse(expiry)
 
         pp = pp_objs[bsuid] = Position(
-            Symbol.from_fqsn(fqsn, info={}),
+            Symbol.from_fqsn(
+                fqsn, entry['symbol']),
```
goodboy, line 985: As mentioned in chat, we need to do a `entry.get('symbol')` here and instead just fill in a default `Symbol` here to continue supporting old `pps.toml` files.

Further for entries detected as `crypto` we should probably put a big warning that the entry needs to be recalculated to avoid the precision error this patch was originally meant to fix!

goodboy (changes requested):
I added in the bit about backward compat for existing `pps.toml` files (since this definitely breaks for me 😂 

I'm more then happy to make a patch to this PR before landing 🏄🏼

goodboy:
With the merging of my sub-pr i think this is ready to land 🏄🏼 

i saw one last msg that was interesting with a `-0.0`, not sure what that's about.

```python
 piker.clearing._ems _ems.py:694 Received broker trade event:
{'account': 'paper',
 'avg_price': 22409.6,
 'broker': 'binance',
 'currency': 'btcusdt',
 'name': 'position',
 'size': -0.0,
 'symbol': 'btcusdt.binance'}
 ```
 
 gonna do one more review pass.

goodboy (approved):
Pretty sure this is finally ready!

I guess we just need to make sure the tests are all good in CI 🚀

PR #473, opened 2023-02-28
`binance`: use built-in `anext()` add note about new ws ep URL, fix agen streaming within `NoBsWs` usage

goodboy:
Not sure why this ain't workin for me (but maybe someone else can test
to verify?) but I tried updating the latest documented WS endpoint and
it doesn't seem to ever work 😂 

Further this starts using the new built-in `anext()` routines instead of
the dunders we had to use prior.

---
**ALSO** ~updates the streamer async-gen with an `aclosing()` which is
super critical to avoid `trio` crashes on re-connection logic!~

**INSTEAD** uses `@trio_util.trio_async_generator` which *actually*
solves the underlying issue of allowing an agen to work through multiple
cancel scopes as is needed by our `NoBsWs` reconnection logic.. **=> in [609b91e](https://github.com/pikers/piker/pull/473/commits/609b91e848740ab92a41a2666d1ca2c2f3c9ef9a)**

### DIFF
--- a/piker/brokers/binance.py
+++ b/piker/brokers/binance.py
@@ -531,17 +531,19 @@ async def stream_quotes(
 
         async with open_autorecon_ws(
             'wss://stream.binance.com/ws',
+            # XXX: see api docs which show diff addr?
+            # https://developers.binance.com/docs/binance-trading-api/websocket_api#general-api-information
+            # 'wss://ws-api.binance.com:443/ws-api/v3',
             fixture=subscribe,
         ) as ws:
 
             # pull a first quote and deliver
             msg_gen = stream_messages(ws)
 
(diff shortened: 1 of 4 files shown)


goodboy:
Ahh shoot, i guess you have to checkout and uncomment the other ep url line 😂 

here's their docs on this: https://developers.binance.com/docs/binance-trading-api/websocket_api#general-api-information

no idea what's up, but obviously as per this patch, we were using `'wss://stream.binance.com/ws'`

jaredgoldman (approved):
lgtm

goodboy:
Turns out this didn't end up workin that well 😂 

Had to do a full rewrite in 59743b7b736957596@#489

PR #474, opened 2023-02-28
`ib`: restore and (maybe) use `xdotool` + `i3ipc` reset method

goodboy:
Since apparently the container we were using is totally borked on new kernels and/or latest jvm, this move our old manual local-X-desktop script back for use in `brokerd` backend code.

Adds a new `.brokers.ib._util` which contains the 2 methods and fails over to this one when we can't connect to a VNC server. Also adjusts the original in `scripts/ib_data_reset.py` to import and run the module code as a script-program.

NB: as i write this i of course write this i realize why we originally "deprecated" this tech 😂 (mouse hijacked during normal desktop usage all the time)

### DIFF
--- a/piker/brokers/ib/_util.py
+++ b/piker/brokers/ib/_util.py
@@ -1,0 +1,184 @@
+# piker: trading gear for hackers
+# Copyright (C) Tyler Goodlet (in stewardship for pikers)
+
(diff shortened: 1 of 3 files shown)


goodboy:
Probably just gonna land this without review since noone else can really test it and no code actually changed, just got moved around.

PR #475, opened 2023-03-01
Explicitly write `pps.toml` on exit for `ib` and `kraken`

goodboy:
As per the default setting change to `.pp.open_pps()` to not write the TOML file on close, this adjusts the `ib` and `kraken` backends to set the write flag explicitly and indeed update the pps file on teardown.

Further, this includes the removal of the trades history record-limit since there seems to be no particular reason to halt on the last 10 transactions, though eventually we should probably expose this through the `brokers.toml`?

### DIFF
--- a/piker/brokers/kraken/broker.py
+++ b/piker/brokers/kraken/broker.py
@@ -470,6 +470,7 @@ async def trades_dialogue(
             open_pps(
                 'kraken',
                 acctid,
+                write_on_exit=True,
             ) as table,
 
             open_trade_ledger(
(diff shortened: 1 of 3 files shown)


jaredgoldman (approved):
Lurkin gurd buddy

piker/brokers/kraken/broker.py
```diff
@@ -1212,23 +1213,3 @@ def norm_trade_records(
         )
 
     return records
-
-
-@cm
-def open_ledger(
```
goodboy, line 1218: oh right, and i have no idea what the heck this was from lul

Issue #467, opened 2023-02-24
`FQM[N/S]`: 'fully qualified market name's, the thing no suit could imagine..


goodboy:
This is a major design decision that seems to have not been solved yet by any other platform (which is frankly baffling): *what does a fully qualified market addressing scheme look like?*

----
#### technical jargon/terminology
Here are 2 options (please add more) for the acronym which describes the url-like address:
- `fqsn`: fully qualified symbol name
- `fqmn`: fully qualified market name (my preference i think?)
- `fqan`: .. auction name
- `fqma`: .. market address
...

----
#### Syntax, labelling and semantics:
There are generally 2 ordering schemes we're thinking are best, both somewhat based on tree-like hierarchical design of [DNS](https://en.wikipedia.org/wiki/Domain_Name_System#Domain_name_syntax,_internationalization) and more generally on [URL style syntax](https://en.wikipedia.org/wiki/URL#Syntax).

resolver op char:
- `.` seems like the most obvious but i think we should be open to `:`
- are there any better chars that are less susceptible to the parsers of shells and other copy/paste from/to envs that will likely be used for passing around the address strings?

current hierarchy-label order:
- `<marketpairname>.<venue>.<suffix(es) with metadata>.brokerbackendname>`
  - eg. `mnqusd.cme.20230317.ib`
  - this obviously becomes more confusing/cumbersome with option market addresses and any other or future added derivs market meta-info.
  
preferred non-reverse hierarchy syntax:
- `ib.cme.mnqusd.20230317.<further option mkt address labels>`
- `<broker>.<venue>.<underlying market pair>.<expiry info label (derivs only)>.<strike price info (opts only)>.<more?>`

I personally much prefer this second (order) style due to it looking more like *attribute access** and will make a UI representation in the search results tree orient from left-to-right exactly the same as scoped search results:

```
ib
|_ cme
     |_ mnqusd
         |_20230317  (expiry price)
             |_1200c
```
which also makes showing the specificity of the market graphically somewhat intuitive?

---
#### Previous work, research, and lead-up to this:
From our original MVP PR https://github.com/pikers/piker/pull/295:
> Basically the idea here is for all client-ish code (aka programs/actors that ask the provider agnostic layers in the stack for data) should be able to tell which backend / venue / derivative each data feed/flow is from by an explicit string key of the current form <instrumentname>.<venue>.<suffixwithmetadata>.<brokerbackendname>.

[...]

PR #477, opened 2023-03-03
Backward compat support for `Transaction.sym: Symbol`

goodboy:
Like it sounds makes #470 still work on existing `pps.toml` files in the wild 😉 

Further this ports the `kraken` and `ib` backends to correctly provide the precision info for transaction sizes where possible by pulling the info from the market info available either from ledger or API records.

---
Additionally, this includes a draft implementation for a `.data._source.Symbol` replacement: `MktPair`.

It's not used at all but gives us a guide for how we could majorly simplify the interface and provide a much clearer defined market addressing system especially as it pertains to refining our schema as documented in #467.

### DIFF
--- a/piker/data/_source.py
+++ b/piker/data/_source.py
@@ -18,7 +18,10 @@
 numpy data source coversion helpers.
 """
 from __future__ import annotations
-from decimal import Decimal, ROUND_HALF_EVEN
+from decimal import (
+    Decimal,
+    ROUND_HALF_EVEN,
+)
 from typing import Any
 
 from bidict import bidict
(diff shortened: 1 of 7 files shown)


piker/data/_source.py
```diff
@@ -127,6 +134,56 @@ def unpack_fqsn(fqsn: str) -> tuple[str, str, str]:
     )
 
 
+class MktPair(Struct, frozen=True):
```
goodboy, line 137: Wrt this, i think if we want to have multiple endpoint brokers addressable eventually we can have type that maybe groups this type into some sort of address collection of sorts?

guilledk (approved):

PR #483, opened 2023-03-08
`.service` subpkg

goodboy:
Move all actor-service control APIs and the service manager (basically all the stuff that was in `piker._daemon`) into a new `piker.service` sub package; this will *serve* (get it..) as the high-level actor-service orchestration sub-system, the business logic for `piker`'s distributed architecture and runtime deployment.

---
#### Of note,
- includes improvements to `.service._ahab` which repair a regression: a silent startup failure with `marketstore` due to the log processing poll sleep period change..
  - adds proper async log msg proxying **without blocking** the `ahabd` supervisor task
  - proxies through `pikerd` logging correctly to the docker super code in general..
- [x] moved the `_ahab.py` docker supervisor into the sub-package as well since it is literally an external, containerized service supervisor 😂 
- [x] moved both the `.data.elastix` and `.marketstore` mods here as well until we figure out how to formalize our *storage* layer subsystems, and since most of that code is related to db interaction/mangement vs. data processing.

---
#### ToDo:
- [ ] ~extend the tests for both dbs to have a client connect and do something or other..~
- [ ] ~add cancellation tests for the dockerized  services~
  **=> moved these to a follow up testing task #487**
- [x] consider breaking out the daemon endpoints to a `._daemon.py` module in each sub-system package?
  - [x] at the least move the group into a `.service._daemon`
  - [ ] ~eg. `pikerd.data._daemon` contains the routines for spawning the `brokerd` actor?~ **=> moved to  #488**
 - [x] move `Services` api into it's own module?
 - [x] move the startup machinery into `_runtime.py`? (** chose `.service._actor_runtime.py` instead **)
 - [ ] ~move the `._exec.py` stuff (which is more or less just Qt runtime startup) into this pkg as well?~ **=> also moved to  #488**
- [x] make test harness use [the `tmp_dir` fixure](https://docs.pytest.org/en/6.2.x/tmpdir.html#the-tmp-path-fixture) for our config dir path during testing
  - required augmenting the runtime init to passthrough data to `tractor._state._runtime_vars`


piker/data/cli.py
```diff
@@ -137,18 +136,74 @@ def storesh(
...
+            'tsdb_storage',
+            enable_modules=['piker.service._ahab'],
         ):
             symbol = symbols[0]
-            await tsdb_history_update(symbol)
+            async with open_tsdb_client(symbol) as storage:
+                if delete:
+                    for fqsn in symbols:
+                        syms = await storage.client.list_symbols()
+                        breakpoint()
```
jaredgoldman, line 204: left in a breakpoint unintentionally?

piker/service/_ahab.py
```diff
@@ -322,37 +354,94 @@ async def open_ahabd(
         ) = ep_func(client)
         cntr = Container(dcntr)
 
-        with trio.move_on_after(start_timeout):
-            found = await cntr.process_logs_until(start_lambda)
+        conf: ChainMap[str, Any] = ChainMap(
```
jaredgoldman, line 357: Just curious, why use a ChainMap here?

piker/data/cli.py
goodboy: ish, was WIP, fixed now (ish) via 6a0ae585

piker/service/_ahab.py
goodboy: Because it's easier then writing a lot of `dict.get(blah, default)` stuff and makes everything extensible for if we ever need to offer overrides of container configs for specific use cases..

piker/service/marketstore.py
```diff
@@ -135,7 +135,7 @@ def start_marketstore(
 
     # create dirs when dne
     if not os.path.isdir(config._config_dir):
-       Path(config._config_dir).mkdir(parents=True, exist_ok=True)
+        Path(config._config_dir).mkdir(parents=True, exist_ok=True)
```
jaredgoldman, line 138: ws?

piker/service/marketstore.py
```diff
@@ -659,7 +662,7 @@ async def tsdb_history_update(
     #     - https://github.com/pikers/piker/issues/98
     #
     profiler = Profiler(
-        disabled=False,  # not pg_profile_enabled(),
+        disabled=True,  # not pg_profile_enabled(),
```
jaredgoldman, line 665: is default true on the Profiler purposeful?

jaredgoldman (approved):
lgtm once ci is passing bruddr

goodboy:
nothing like hanging for 40 mins

goodboy:
THERE. moved all oustanding bullets to new follow up issues 🏄🏼 
- #487
- #488

So whoever decides to review this once more and decides we gud, please merge.
If not i'll land later tonight 💥

goodboy:
Heh, i can tell y'all aren't testing non-disti mode.. 😂

Issue #485, opened 2023-03-09
tsdb multi-backend support and testing!


goodboy:
As per the slew of issues pertaining to time-series-db subsystems in piker seen here:
https://github.com/pikers/piker/labels/tsdb

---
#### Hi-level ToDo:
- [ ] breaking out the middle-ware history managment layer into it's own subpkg as per https://github.com/pikers/piker/issues/315
- [ ] trying out **at least one other** tsdb (probably `arctic`) and formalizing our OHLCV history apis as per the research options listed in #90
- [ ] actually documenting how to launch the db with `--tsdb` 😂 as per #313 
- [ ] generally speaking, much better management tools for manually pulling data from providers and populating a db instance either with cli or context-menu options in our charting app!
- [ ] real-time ingest of tick and signals (from our fsp layer) as per, for eg. #314
- [ ] a tick history to 1s OHLCV downsampler system that can be used cross-provider since most don't provide this low of a sample rate (think only one is `ib`?)

...

maybe more?

---
#### testing ToDo:
- [ ] db client connection, write, read and delete a time series loaded from either an offline sample or some provider we can always rely on (like `kraken` maybe?)
- [ ] basic tests for replication for those dbs which support it

PR #486, opened 2023-03-09
Start `piker.storage` subsys: cross-(ts)db middlewares

goodboy:
Launch pad for work towards the task list in #485 🏄🏼 

As a start this introduces a new `piker.storage` subsystem to provide database related middlewares.

---
#### ToDo:
- see #485
- [ ] probably more specific tasks 😂


goodboy:
We can convert this to a draft if necessary if/when #483 lands

guilledk:
I'm in favor of doing or own solution and I would rather stop maintaining any marketstore related coded, in the end we were almost gonna spend as much work mantaining marketstore that just doing our own thing right.

goodboy:
> I'm in favor of doing or own solution and I would rather stop maintaining any marketstore related coded, in the end we were almost gonna spend as much work mantaining marketstore that just doing our own thing right.

yup totally agree!

ok then i'll be putting up some finishing functionality touches, hopefully tests, and then dropping all that junk 🏄🏼

goodboy:
To give an idea of what the parquet subdir looks like now, much in similarity to how `marketstore` laid out it's own internal per table binary format files except using less space and actually being a file type data people can use 😂 
screenshot-2023-06-14_14-08-23

PR #489, opened 2023-03-10
Rekt pps? problem? => `piker.accounting`

goodboy:
Even *moar* fixes after #470, #475, #484 which mega broke the old
`pps.toml` management for live backend 💥 😶‍🌫️ 🤕 

Turns out this might be **the reason** to start assaulting the revamp of
our accounting tooling as per our now long list of desirables:
https://github.com/pikers/piker/labels/accounting

### Testing enhancements (some tasks deferred)
- [x] rework test set to use `pytest.mark.parametrize` so we can more
  easily use common harness logic for darks and alert testing.

- [x] reorg the ahab API docker supervisor endpoints to be decoupled
  from `open_pikerd()` and able to accept arbitrary config overrides for
  both testing and distributed setups.

- [ ] ~~dark order and alert clearing~~
  - #466 and what we're planning to do for offline storage of level
    lines

- [ ] ~~add a basic accounting of position only test set?~~
  - **deferred to #512 testing section**

#### tasks and bugs getting deferred..
- [ ] pre-existing live order loading is borked on first chart edit from
  order mode: https://github.com/pikers/piker/issues/502

- [ ] final removal of `Symbol` and `.symbol` semantics -> summarized in #517
  - [ ] drop from clearing msg set.
  - [ ] `order_mode` `.symbol` final usage..?
- [ ] make a follow up issue to add a `Flume.Cascade` type for fsps
  which describes stream IO between flumes:
- [ ] deribit still need full `get_mkt_info()` and `FeedInit`
- [x] sync up the new `data.validate` stuff with old brokercheck idea in
  #387, #393 -> add an actual test?

#### dependency changes needed for this to land

- [x] move to new `TOML` libs (`tomlkit` for styling and `tomli-w` for
      fast writing) B)
  - [x] break the shoddy `toml` writing hacks we have into a new submod..
  - [x] actually remove the `toml` dep and go with `tomli`, `tomli-w`
    and `tomlkit` B)
  - [x] move to using `tomlkit` which **does support** writing inline
    tables (as well as lotsa other great stuff) as requested in #496 B)
    - [x] For now **pin to our fork**!
      (https://github.com/pikers/tomlkit/tree/piker_pin)
      - need to land all local changes and pin the requirement!
    - [x] make PR to upstream -> https://github.com/sdispater/tomlkit/pull/294

- [ ] https://github.com/goodboy/tractor/pull/357
  - [x] at least move the piker pin to this history to get CI green

---
### Unrelated to `.accounting` changes
[...]

PR #494, opened 2023-03-19
kucoin backend

jaredgoldman:
This PR includes a brokerd module for [Kucoin](https://www.kucoin.com)'s [api](https://www.kucoin.com/api). So far, only code for market data has been implemented. This includes:

- Fetching a users historical trades
- Subscribing to real-time market data via the Kucoin websocket feed
- Multi-chartz support
- Fetching OHLC historical market data
- Kucoin user auth
- Symbol fuzzy search
---
Remaining tasks for full support are detailed in [this issue](https://github.com/pikers/piker/issues/499)


goodboy:
@jaredgoldman this ready for serious review?

jaredgoldman:
@goodboy I think we're ready for some more testing

goodboy:
@algorandpa seems like i can only stage order with `size=0` but other then that historical and live data seems right 🏄🏼

piker/brokers/kucoin.py
```diff
@@ -0,0 +1,668 @@
...
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+'''
+Kucoin broker backend
+
+'''
+
+from random import randint
+from typing import Any, Callable, Optional, Literal, AsyncGenerator
```
goodboy, line 22: as per other code,i prefer the multi-line version:

```python
from typing import (
	Any,
	Callable,
	Optional,
	Literal,
	AsyncGenerator,
)
```

piker/brokers/kucoin.py
```diff
@@ -0,0 +1,668 @@
...
+from piker.pp import config
+from ..data.types import Struct
+from ..data._web_bs import (
+    open_autorecon_ws,
+    NoBsWs,
+)
+
+log = get_logger(__name__)
+
+_ohlc_dtype = [
```
goodboy, line 53: don't think you need to define this unless it's diff from our default.

piker/brokers/kucoin.py
```diff
@@ -0,0 +1,668 @@
...
+    ('bar_wap', float),  # will be zeroed by sampler if not filled
+]
+
+
+class KucoinMktPair(Struct, frozen=True):
+    '''
+    Kucoin's pair format
+
+    '''
+
```
goodboy, line 70: i tend to *not* prefer having an extra line between `'''` and docstring since we already put one as last content line in doc string.

piker/brokers/kucoin.py
```diff
@@ -0,0 +1,668 @@
...
+    ('open', float),
+    ('high', float),
+    ('low', float),
+    ('close', float),
+    ('volume', float),
+    ('bar_wap', float),  # will be zeroed by sampler if not filled
+]
+
+
+class KucoinMktPair(Struct, frozen=True):
```
goodboy, line 65: put a comment with a link to api docs for this.

piker/brokers/kucoin.py
```diff
@@ -0,0 +1,668 @@
...
+    priceIncrement: float
+    priceLimitRate: float
+    quoteCurrency: str
+    quoteIncrement: float
+    quoteMaxSize: float
+    quoteMinSize: float
+    symbol: str
+
+
+class AccountTrade(Struct, frozen=True):
```
goodboy, line 90: Again, links to API docs for all these structs.

piker/brokers/kucoin.py
```diff
@@ -0,0 +1,668 @@
...
+        return None
+
+    return BrokerConfig(**section)
+
+
+class Client:
+    def __init__(self) -> None:
+        self._pairs: dict[str, KucoinMktPair] = {}
+        self._bars: list[list[float]] = []
+        self._key_id: str
```
goodboy, line 172: are these duplicate with the fields of `BrokerConfig`?
If so why do you need both sets?

goodboy: just do `self._config: BrokerConfig` no?

fwiw `msgspec.Struct`s don't do any validation, so packing it in one without doing anything else is basically a noop.

piker/brokers/kucoin.py
```diff
@@ -0,0 +1,668 @@
...
+        self._key_secret: str
+        self._key_passphrase: str
+        self._authenticated: bool = False
+
+        config: BrokerConfig | None = get_config()
+
+        if config and config.key_id and config.key_secret and config.key_passphrase:
```
goodboy, line 179: as per above. either choose to wrap creds in a config var or do it in init method here; you don't need both.

piker/brokers/kucoin.py
```diff
@@ -0,0 +1,668 @@
...
+        action: Literal['POST', 'GET'],
+        endpoint: str,
+        api_v: str = 'v2',
+        headers: dict = {},
+    ) -> Any:
+        '''
+        Generic request wrapper for Kucoin API
+
+        '''
+        if self._authenticated:
```
goodboy, line 237: shouldn't this be the inverse?

only regen auth headers when *not* already authed?

piker/brokers/kucoin.py
```diff
@@ -0,0 +1,668 @@
...
+
+        if 'data' in res.json():
+            return res.json()['data']
+        else:
+            log.error(f'Error making request to {api_url} -> {res.json()["msg"]}')
+            return res.json()['msg']
+
+    async def _get_ws_token(self, private: bool = False) -> tuple[str, int] | None:
+        '''
+        Fetch ws token needed for sub access
```
goodboy, line 252: api docs link

piker/brokers/kucoin.py
```diff
@@ -0,0 +1,668 @@
...
+        '''
+        token_type = 'private' if private else 'public'
+        data: dict[str, Any] | None = await self._request(
+            'POST', f'/bullet-{token_type}', 'v1'
+        )
+
+        if data and 'token' in data:
+            ping_interval: int = data['instanceServers'][0]['pingInterval']
+            return data['token'], ping_interval
+        elif data:
```
goodboy, line 263: what's normally happening in this case?

piker/brokers/kucoin.py
```diff
@@ -0,0 +1,668 @@
...
+            log.error(
+                f'Error making request for Kucoin ws token -> {data.json()["msg"]}'
+            )
+
+    async def _get_pairs(
+        self,
+    ) -> dict[str, KucoinMktPair]:
+        if self._pairs:
```
goodboy, line 271: wait, why?

piker/brokers/kucoin.py
```diff
@@ -0,0 +1,668 @@
...
+
+        entries = await self._request('GET', '/symbols')
+        syms = {item['name']: KucoinMktPair(**item) for item in entries}
+
+        log.info('Kucoin market pairs fetches')
+        return syms
+
+    async def cache_pairs(
+        self,
+        normalize: bool = True,
```
goodboy, line 282: no, do not mutate state and store is based on a flag 😂

piker/brokers/kucoin.py
```diff
@@ -0,0 +1,668 @@
...
+        Get cached pairs and convert keyed symbols into fqsns if ya want
+
+        '''
+        if not self._pairs:
+            self._pairs = await self._get_pairs()
+        if normalize:
+            self._pairs = self._normalize_pairs(self._pairs)
+        return self._pairs
+
+    def _normalize_pairs(
```
goodboy, line 294: you don't need both of these methods. just pick the one with the the normalized output, unless there's some reason not to?

piker/brokers/kucoin.py
```diff
@@ -0,0 +1,668 @@
...
+        limit: int = 30,
+    ) -> dict[str, KucoinMktPair]:
+        data = await self._get_pairs()
+
+        matches = fuzzy.extractBests(pattern, data, score_cutoff=35, limit=limit)
+        # repack in dict form
+        return {kucoin_sym_to_fqsn(item[0].name): item[0] for item in matches}
+
+    async def last_trades(self, sym: str) -> list[AccountTrade]:
+        trades = await self._request('GET', f'/accounts/ledgers?currency={sym}', 'v1')
```
goodboy, line 321: link to api docs.

goodboy:
@jaredgoldman as mentioned in chat as well we've got some kinda strange thing going on with trade rate calcs:
- there's some kind of trade event being picked up by the fsp system 10/sec
- you can see flat curve for this if you leave it running
- this **does not match up with** actual vlm / clearing rates

screenshot-2023-03-29_21-32-25

piker/brokers/kucoin.py
```diff
@@ -0,0 +1,668 @@
...
+                task_status.started((init_msgs, quote))
+                feed_is_live.set()
+
+                async for typ, msg in msg_gen:
+                    await send_chan.send({sym: msg})
+
+
+def make_sub(sym, connect_id, level='l1') -> dict[str, str | bool]:
+    match level:
+        case 'l1':
```
goodboy, line 541: Yeah so re my comment about the trade rates mis-calculation, I think this might be the cause.

maybe one of these subs is being reported every 10ms and you're not filtering / diffing for actual updates since the last msg?

likely you'll need to keep a local book state and only transmit to piker when there's a real update to the book and/or real clear events that have taken place.

goodboy: yeah comment is just above https://github.com/pikers/piker/pull/494#issuecomment-1493068034

piker/brokers/kucoin.py
```diff
@@ -0,0 +1,668 @@
...
+            pairs = await client.cache_pairs()
+            kucoin_sym = pairs[sym].symbol
+
+            init_msgs = {
+                # pass back token, and bool, signalling if we're the writer
+                # and that history has been written
+                sym: {
+                    'symbol_info': {
+                        'asset_type': 'crypto',
+                        'price_tick_size': 0.0005,
```
goodboy, line 464: wait why are you hardcoding this instead of pulling it from the pair info in `pairs`?

piker/brokers/kucoin.py
```diff
@@ -0,0 +1,668 @@
...
+                },
+            }
+
+            @acm
+            async def subscribe(ws: wsproto.WSConnection):
+                @acm
+                async def open_ping_task(ws: wsproto.WSConnection):
+                    '''
+                    Spawn a non-blocking task that pings the ws
+                    server every ping_interval so Kucoin doesn't drop
```
goodboy, line 478: what's `ping_interval`?

piker/brokers/kucoin.py
jaredgoldman: Done

piker/brokers/kucoin.py
jaredgoldman: Done

piker/brokers/kucoin.py
```diff
@@ -0,0 +1,668 @@
...
+        api_v: str = 'v2',
+    ) -> dict[str, str | bytes]:
+        '''
+        Generate authenticated request headers
+        https://docs.kucoin.com/#authentication
+
+        '''
+        now = int(time.time() * 1000)
+        path = f'/api/{api_v}{endpoint}'
+        str_to_sign = str(now) + action + path
```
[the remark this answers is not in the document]
jaredgoldman: slick

piker/brokers/kucoin.py
jaredgoldman: Let me test this - I assumed since the headers need a timestamp that you'd have to create new ones for each request. In retrospect I could totally be wrong about this and haven't tested to evaluate my assumptions.

jaredgoldman: Ah but wait, we have to pass the HTTP action into the header... @goodboy we have to encode to action and timestamp into the headers like so:
```py
str_to_sign = str(int(time.time() * 1000)) + action + f'/api/{api_v}{endpoint}'

        signature = base64.b64encode(
            hmac.new(
                self._key_secret.encode('utf-8'),
                str_to_sign.encode('utf-8'),
                hashlib.sha256,
            ).digest()
        )
```
Worth it to try to generate some of the headers ahead of time?

piker/brokers/kucoin.py
jaredgoldman: Whoops.

piker/brokers/kucoin.py
jaredgoldman: I'm actually unsure of what the `price_tick_size` and `lot_tick_size` are. Do they map with any values in `KucoinMktPair`?

jaredgoldman:
@goodboy merged the binance changes and add anext()

piker/brokers/kucoin.py
jaredgoldman: @goodboy - I think we need better documentation on init message formatting.

piker/brokers/kucoin.py
jaredgoldman: check out the docs in the docstring for `_get_ws_token`

piker/brokers/kucoin.py
goodboy: yah, but i think if you look, the default schema def is applied if not defined..

it should only be set if something is diff from the default iirc

we should probably document this.

piker/brokers/kucoin.py
```diff
@@ -0,0 +1,670 @@
...
+def get_config() -> BrokerConfig | None:
+    conf, _= config.load()
+
+    section = conf.get('kucoin')
+
+    if section is None:
+        log.warning('No config section found for kucoin in config')
+        return None
+
+    return BrokerConfig(**section)
```
goodboy, line 179: i kinda like this. if you wanted to ensure the typing you could call `.typecast()` i think?

https://github.com/pikers/piker/blob/master/piker/data/types.py#L83

piker/brokers/kucoin.py
```diff
@@ -0,0 +1,670 @@
...
+    key_secret: str
+    key_passphrase: str
+
+
+def get_config() -> BrokerConfig | None:
+    conf, _= config.load()
+
+    section = conf.get('kucoin')
+
+    if section is None:
```
goodboy, line 175: what happens in this case?

piker/brokers/kucoin.py
goodboy: so if `bool(data)` being false-y should never happen yah?

cuz in that case this will exit silently i think?

piker/brokers/kucoin.py
```diff
@@ -0,0 +1,670 @@
...
+        elif data:
+            log.error(
+                f'Error making request for Kucoin ws token -> {data.json()["msg"]}'
+            )
+
+    async def _get_pairs(
+        self,
+    ) -> dict[str, KucoinMktPair]:
+        entries = await self._request('GET', '/symbols')
+        syms = {kucoin_sym_to_fqsn(item['name']): KucoinMktPair(**item) for item in entries}
```
goodboy, line 278: if you want you can be all hip and go with the new `fqme` from #489 😂

piker/brokers/kucoin.py
```diff
@@ -0,0 +1,670 @@
...
+
+        log.info('Kucoin market pairs fetched')
```
goodboy, line 280: maybe toss the lenth of `syms` in this msg?

probably something we should do on all backends too

piker/brokers/kucoin.py
```diff
@@ -0,0 +1,670 @@
...
+        return syms
+
+    async def cache_pairs(
+        self,
+        # normalize: bool = True,
```
goodboy, line 285: drop this?

piker/brokers/kucoin.py
```diff
@@ -0,0 +1,670 @@
...
+    ) -> dict[str, KucoinMktPair]:
+        '''
+        Get cached pairs and convert keyed symbols into fqsns if ya want
+
+        '''
+        if not self._pairs:
+            self._pairs = await self._get_pairs()
```
goodboy, line 292: i think it'd be cool if in the longer run we actually cache these somehow to local storage per backend..

just thinking out loud.

piker/brokers/kucoin.py
```diff
@@ -0,0 +1,670 @@
...
+
+        return self._pairs
+
+    async def search_symbols(
+        self,
+        pattern: str,
+        limit: int = 30,
+    ) -> dict[str, KucoinMktPair]:
+        data = await self._get_pairs()
```
goodboy, line 301: wouldn't it make more sense to use `.cache_pairs()` for this call so you aren't constantly reloading?

piker/brokers/kucoin.py
```diff
@@ -0,0 +1,670 @@
...
+
+    async def last_trades(self, sym: str) -> list[AccountTrade]:
+        trades = await self._request('GET', f'/accounts/ledgers?currency={sym}', 'v1')
+        trades = AccountResponse(**trades)
+        return trades.items
+
+    async def _get_bars(
+        self,
+        fqsn: str,
+        start_dt: Optional[datetime] = None,
```
goodboy, line 316: much prefer the new `datetime | None` syntax from 3.10

piker/brokers/kucoin.py
```diff
@@ -0,0 +1,670 @@
...
+        if start_dt is None:
+            start_dt = end_dt.start_of('minute').subtract(minutes=limit)
+
+        start_dt = int(start_dt.timestamp())
+        end_dt = int(end_dt.timestamp())
+
+        kucoin_sym = fqsn_to_kucoin_sym(fqsn, self._pairs)
+
+        url = f'/market/candles?type={type}&symbol={kucoin_sym}&startAt={start_dt}&endAt={end_dt}'
+        bars = []
```
goodboy, line 341: type annot here on the list

piker/brokers/kucoin.py
```diff
@@ -0,0 +1,670 @@
...
+        for i in range(10):
+            data = await self._request(
+                'GET',
+                url,
+                api_v='v1',
+            )
+
+            if not isinstance(data, list):
+                # Do a gradual backoff if Kucoin is rate limiting us
+                backoff_interval = i
```
goodboy, line 352: so this is an `int` from the `range` iter but you're using it as a float to `trio.sleep()` yah?

piker/brokers/kucoin.py
```diff
@@ -0,0 +1,670 @@
...
+                log.warn(
+                    f'History call failed, backing off for {backoff_interval}s')
+                await trio.sleep(backoff_interval)
+            else:
+                bars = data
```
goodboy, line 357: considering you're overriding `bars: list[list]` each backoff algo case, seems weird you pre-define it above as an empty list.

maybe just make the pre-def a type annot only?

piker/brokers/kucoin.py
```diff
@@ -0,0 +1,670 @@
...
+                'close': bar[2],
+                'high': bar[3],
+                'low': bar[4],
+                'volume': bar[5],
+                'amount': bar[6],
+                'bar_wap': 0.0,
+            }
+
+            row = []
+            for _, (field_name, field_type) in enumerate(_ohlc_dtype):
```
goodboy, line 376: hmm, this might be one of the cases where this fancy loop through the hard coded dict you have above might just be better done as a hardcoding of each field in the ohlc array to the respective index in the "bar data" array rx-ed from their api?

piker/brokers/kucoin.py
```diff
@@ -0,0 +1,670 @@
...
+                value = data[field_name]
+
+                match field_name:
```
goodboy, line 379: yeah i think in general this whole "fill loop" stuff is probably just better done by hardcoding the array fields to our numpy array fields manually, and just link to their docs.

it's some interesting gymnastics tho 😂

piker/brokers/kucoin.py
```diff
@@ -0,0 +1,670 @@
...
+) -> None:
+    async with open_cached_client('kucoin') as client:
+        # load all symbols locally for fast search
+        await client.cache_pairs()
+        await ctx.started()
+
+        async with ctx.open_stream() as stream:
+            async for pattern in stream:
+                await stream.send(await client.search_symbols(pattern))
+                log.info('Kucoin symbol search opened')
```
goodboy, line 422: maybe move this above?

piker/brokers/kucoin.py
```diff
@@ -0,0 +1,670 @@
...
+    Required piker api to stream real-time data.
+    Where the rubber hits the road baby
+
+    '''
+    connect_id = str(uuid4())
+
+    async with open_cached_client('kucoin') as client:
+        log.info('Starting up quote stream')
+        # loop through symbols and sub to feedz
+        for sym in symbols:
```
goodboy, line 443: wait you're making a new ws per sym?

piker/brokers/kucoin.py
goodboy: the short answer is yes but also this is being more formalized in #489 .. unfortunately a bit later then i would have liked 😂 

if you wanted to get ambitious you could try to adopt the stuff in there, but it's probably fine to get changed later.

more or less `price/lot_tick_size` are the float/Decimal values which specify the *min increment value* for both the market's price and volume/size.

goodboy: so yes, you should pull them from `KucoinMktPair`.

likely the two values inline here:
https://github.com/pikers/piker/pull/494/files#diff-0e687c4074ef81501ccba2717591ca55647c53b9b3dc85c33d4b4fc1436c644cR78

piker/brokers/kucoin.py
```diff
@@ -0,0 +1,670 @@
...
+                async def open_ping_task(ws: wsproto.WSConnection):
+                    '''
+                    Spawn a non-blocking task that pings the ws
+                    server every ping_interval so Kucoin doesn't drop
+                    our connection
+
+                    '''
+                    async with trio.open_nursery() as n:
+                        # TODO: cache this task so it's only called once
+                        async def ping_server():
```
goodboy, line 473: cute `trio` work 🏄🏼

piker/brokers/kucoin.py
```diff
@@ -0,0 +1,670 @@
...
+                        log.info(f'Starting ping task for {sym}')
+                        n.start_soon(ping_server)
+
+                        yield ws
+
+                        n.cancel_scope.cancel()
+
+                # Spawn the ping task here
+                async with open_ping_task(ws) as ws:
+                    tasks = []
```
goodboy, line 487: this seems generally weird to me, the whole `tasks` bit.

i'd say just either add the subs inline without all the `.append()`ing or just generally drop these 2 kinda weird loops.

i got a feeling you can do the subs requests in maybe one send as well?

piker/brokers/kucoin.py
```diff
@@ -0,0 +1,670 @@
...
+                open_autorecon_ws(
+                    f'wss://ws-api-spot.kucoin.com/?token={token}&[connectId={connect_id}]',
+                    fixture=subscribe,
+                ) as ws,
+                stream_messages(ws, sym) as msg_gen,
+            ):
+                typ, quote = await anext(msg_gen)
+
+                while typ != 'trade':
+                    # TODO: use ``anext()`` when it lands in 3.10!
```
goodboy, line 521: you can drop this comment now, also maybe put a comment about how we don't unblock until we get a first "trade quote"

piker/brokers/kucoin.py
```diff
@@ -0,0 +1,670 @@
...
+                    typ, quote = await anext(msg_gen)
+
+
+                task_status.started((init_msgs, quote))
+                feed_is_live.set()
```
goodboy, line 526: hmm iirc you can actually `.started()` sooner if possible. it's only important to wait on a "real trade quote" (as mentioned above) before setting the `feed_is_live: trio.Event()` i think?

piker/brokers/kucoin.py
```diff
@@ -0,0 +1,670 @@
...
+
+                async for typ, msg in msg_gen:
+                    await send_chan.send({sym: msg})
+
+
+def make_sub(sym, connect_id, level='l1') -> dict[str, str | bool] | None:
```
goodboy, line 532: yeah imo way too many layers for all this subs stuff.

Just render the subs dicts inline in the `subscribe()` func above.

piker/brokers/kucoin.py
```diff
@@ -0,0 +1,670 @@
...
+                'id': connect_id,
+                'type': 'subscribe',
+                'topic': f'/market/ticker:{sym}',
+                'privateChannel': False,
+                'response': True,
+            }
+
+@trio_async_generator
+async def stream_messages(ws: NoBsWs, sym: str) -> AsyncGenerator[NoBsWs, dict]:
+    timeouts = 0
```
goodboy, line 554: type annots are nice 😉

piker/brokers/kucoin.py
```diff
@@ -0,0 +1,670 @@
...
+        if cs.cancelled_caught:
+            timeouts += 1
+            if timeouts > 2:
+                log.error('kucoin feed is sh**ing the bed... rebooting...')
+                await ws._connect()
+
+            continue
+
+
+        if msg.get('subject') != None:
```
goodboy, line 569: you don't need the `!= None` check i don't think?

goodboy: i mean unless you expect a `False`-y value to be returned for this?

piker/brokers/kucoin.py
```diff
@@ -0,0 +1,670 @@
...
+                    yield 'trade', {
+                        'symbol': sym,
+                        'last': trade_data.price,
+                        'brokerd_ts': trade_data.time,
+                        'ticks': [
+                            {
+                                'type': 'trade',
+                                'price': float(trade_data.price),
+                                'size': float(trade_data.size),
+                                'broker_ts': trade_data.time,
```
goodboy, line 591: why not use the `last_trade_ts` var you just set to avoid yet another attr lookup.

piker/brokers/kucoin.py
```diff
@@ -0,0 +1,670 @@
...
+            times = array['time']
+
+            if end_dt is None:
+                inow = round(time.time())
+
+                print(
+                    f'difference in time between load and processing {inow - times[-1]}'
+                )
+
+                if (inow - times[-1]) > 60:
```
goodboy, line 660: heh, you really are copying my janky code 😂

goodboy (changes requested):
@jaredgoldman this is looking super dupe good bud!

I have a few change requests, but i've tested the branch a bit now and it's feeling good!

I think we should maybe also breakout the live trading control / orders stuff into a new issue for follow up. It'd bee good to land this data feed stuff before 😎

piker/brokers/kucoin.py
jaredgoldman: This confuses me. I'm using it to parse my data into the numpy array
```py
            for _, (field_name, field_type) in enumerate(_ohlc_dtype):
                value = data[field_name]

                match field_name:
                    case 'index':
                        row.append(int(value))
                    case 'time':
                        row.append(value)
                    case _:
                        row.append(float(value))

            new_bars.append(tuple(row))
```

piker/brokers/kucoin.py
jaredgoldman: Let's account for that just in case. If the API request is malformed this could happen.

piker/brokers/kucoin.py
jaredgoldman: We do not generate auth headers for requests 
```py
 if self._config:
            headers = self._gen_auth_req_headers(action, endpoint, api_v)
```

piker/brokers/kucoin.py
jaredgoldman: Great idea! Also a good opportunity for maybe a redis service that will periodically refresh it's cache.

piker/brokers/kucoin.py
jaredgoldman: you got it. Very simple logic

piker/brokers/kucoin.py
jaredgoldman: lol I copied it from @guilledk

piker/brokers/kucoin.py
jaredgoldman: Above what

piker/brokers/kucoin.py
jaredgoldman: seems like task started requires a real quote.

piker/brokers/kucoin.py
goodboy: sure moreso my point is i don't think you need to define it at all in the module unless it's diff then the default:
https://github.com/pikers/piker/blob/master/piker/data/_source.py#L34

the logic which decides i believe is here in the data layer:
https://github.com/pikers/piker/blob/master/piker/data/feed.py#L799

so i'm just saying you might not need to even define it and if you need to iterate fields, just iterate the `base_iohlc_dtype`?

further this whole loop thinger:
```python
  for _, (field_name, field_type) in enumerate(_ohlc_dtype):
                value = data[field_name]
```

is this really any better then just doing inline static assignment?

for eg. here's how we do it in the `ib` backend:
https://github.com/pikers/piker/blob/master/piker/brokers/ib/api.py#L271

you'll notice it's just based on ordering the fields according to the client lib/API's output format, and we just kinda inline map it to the default ohlc dtype.

goodboy:
Looking the code in the ol' vim i'm seeing a lot of these brother:

screenshot-2023-04-14_19-03-31

goodboy:
Added patch commit to get things working again after your review changes in #498

piker/brokers/kucoin.py
jaredgoldman: Word, I'll simplify the logic

piker/brokers/kucoin.py
jaredgoldman: Couldn't figure out the multiple requests in one send but I've factored out this abstraction.

piker/brokers/kucoin.py
jaredgoldman: I could move it earlier in the execution flow and then provide a fake quote? @goodboy thoughts?

jaredgoldman:
@goodboy I refactored the `subscribe` and `stream_messages` logic to:
- Only use one ws connection and one `ping_task` 
- Be less nested and more readable

lol, the earlier implementation was wildy inefficient spawning a diff ws connection and ping task for each symbol. In addition I've addressed the comments for which I'm not waiting on a reply from you (no rush). We're also getting to the point where we should consider factoring this into a few different files, especially before adding order control.

piker/brokers/binance.py
```diff
@@ -551,7 +551,7 @@ async def subscribe(ws: wsproto.WSConnection):
             while typ != 'trade':
                 typ, quote = await anext(msg_gen)
 
-            task_status.started((init_msgs,  quote))
+            task_status.started((init_msgs, quote))
```
goodboy, line 554: this is just deleting a space yah?

piker/brokers/kucoin.py
```diff
@@ -0,0 +1,685 @@
...
+) -> AsyncGenerator[Callable, None]:
+    async with open_cached_client('kucoin') as client:
+
+        log.info('Attempting to open kucoin history client')
+
+        async def get_ohlc_history(
+            timeframe: float,
+            end_dt: datetime | None = None,
+            start_dt: datetime | None = None,
+        ) -> tuple[np.ndarray, datetime | None, datetime | None]:  # start  # end
```
goodboy, line 655: i would make this a mult-liner, as i do whenever you have this many diff types in one annotation.

piker/brokers/kucoin.py
```diff
@@ -0,0 +1,685 @@
...
+
+            if end_dt is None:
+                inow = round(time.time())
+
+                print(
+                    f'difference in time between load and processing {inow - times[-1]}'
+                )
+
+                if (inow - times[-1]) > 60:
+                    await tractor.breakpoint()
```
goodboy, line 676: drop this 😂

piker/brokers/kucoin.py
```diff
@@ -0,0 +1,685 @@
...
+            msg = await ws.recv_msg()
+        if cs.cancelled_caught:
+            timeouts += 1
+            if timeouts > 2:
+                log.error('kucoin feed is sh**ing the bed... rebooting...')
+                await ws._connect()
+
+            continue
+
+        if msg.get('subject'):
```
goodboy, line 584: FYI (though not that you have to change it right now) you could actually match against your `msg` and then use object style matching for accessing all the values of the msg if you prefer:

https://jcristharif.com/msgspec/structs.html#pattern-matching

goodboy: ahh just to follow up, so that means you can use the object attributes as both a match condition, **and** an upacker system for binding to local variables in the scope:
https://peps.python.org/pep-0636/#adding-a-ui-matching-objects

piker/brokers/kucoin.py
```diff
@@ -0,0 +1,685 @@
...
+                    task_status.started((init_msgs, quote))
+                    feed_is_live.set()
+
+                    async for typ, msg in msg_gen:
+                        await send_chan.send({sym: msg})
+
+@acm
+async def subscribe(ws: wsproto.WSConnection, connect_id, sym):
+    # level 2 sub
+    await ws.send_msg({
```
goodboy, line 535: i'm surprised you can't do both of these subs in one msg?

piker/brokers/kucoin.py
```diff
@@ -0,0 +1,685 @@
...
+            # loop through symbols and sub to feedz
+            for sym in symbols:
+                pair: KucoinMktPair = pairs[sym]
+                kucoin_sym = pair.symbol
+
+                init_msgs = {
+                    # pass back token, and bool, signalling if we're the writer
+                    # and that history has been written
+                    sym: {
+                        'symbol_info': {
```
goodboy, line 506: i'm happy to change this as part of landing #489 but if you wanted to line up with that you can try to as well.

though i guess that won't make it compat with master.. so maybe not?

piker/brokers/kucoin.py
```diff
@@ -0,0 +1,685 @@
...
+        await client.cache_pairs()
+        await ctx.started()
+
+        async with ctx.open_stream() as stream:
+            async for pattern in stream:
+                await stream.send(await client.search_symbols(pattern))
+                log.info('Kucoin symbol search opened')
+
+@acm
+async def open_ping_task(ws: wsproto.WSConnection, ping_interval, connect_id):
```
goodboy, line 449: missing type annots

piker/brokers/kucoin.py
```diff
@@ -0,0 +1,685 @@
...
+        self,
+        fqsn: str,
+        start_dt: datetime | None = None,
+        end_dt: datetime | None = None,
+        limit: int = 1000,
+        as_np: bool = True,
+        type: str = '1min',
+    ) -> np.ndarray:
+        '''
+        Get OHLC data and convert to numpy array for perffff:
```
goodboy, line 328: bootyful

piker/brokers/kucoin.py
```diff
@@ -0,0 +1,685 @@
...
+        url = f'/market/candles?type={type}&symbol={kucoin_sym}&startAt={start_dt}&endAt={end_dt}'
+
+        for i in range(10):
+            data = await self._request(
+                'GET',
+                url,
+                api_v='v1',
+            )
+
+            if not isinstance(data, list):
```
goodboy, line 377: what is the normal type of `data` ?
would be handy to type annot this even if it is somewhat variable 😉

goodboy (changes requested):
Other then this linting output (which i think i've hassled
screenshot-2023-04-17_18-29-54
 about enough at this point) 😂 
 
 This is **sooo close**.
 
 I have couple more quiffs mostly to do with typing 🥱 
 
 Testing says this is basically ready to land, even playing with the paper clearing engine a bunch as well as overlays 👍🏼

goodboy:
> We're also getting to the point where we should consider factoring this into a few different files, especially before adding order control.

yes, if you want to in this PR it's fine by me. currently we something like:
- core backend api client and general utils in `.api`
- data feed setup, search task and history client in a `.feed`
- all the trade control stuff (which i don't think you have any of yet) in a `.broker`

piker/brokers/kucoin.py
jaredgoldman: Although I do see the value of this, I'm not matching for anything more specific than the subject. I think we're fine to leave this for now.

piker/brokers/kucoin.py
jaredgoldman: Seems like what's even the point the way their API works? https://docs.kucoin.com/#multiplex @goodboy

piker/brokers/kucoin.py
jaredgoldman: Maybe handle it 489 in that case.

piker/brokers/kucoin.py
```diff
@@ -0,0 +1,718 @@
...
+                                'size': float(first_ask[1]),
+                            },
+                        ],
+                    }
+
+
+@acm
+async def open_history_client(
+    symbol: str,
+    type: str = '1m',
```
goodboy, line 678: sorry just noticed this..
try not to use built-ins as var names if you can, also i don't think this is used in the func body?

goodboy: yeah i think you can just drop it no?

`timeframe` info is passed into the delivered closure on call.

piker/brokers/kucoin.py
```diff
@@ -0,0 +1,718 @@
...
+            msg = await ws.recv_msg()
+        if cs.cancelled_caught:
+            timeouts += 1
+            if timeouts > 2:
+                log.error('kucoin feed is sh**ing the bed... rebooting...')
+                await ws._connect()
+
+            continue
+
+        if msg.get('subject'):
```
goodboy, line 616: you *could* do the match on `msg` directly and just do patterns against the `.subject` field, but honestly not a big deal right now; we can polish it up a bit later.

also, you should normally have a `case _:` and report non-matches that aren't really expected 😉

piker/brokers/kucoin.py
```diff
@@ -0,0 +1,718 @@
...
+
+                    task_status.started((init_msgs, quote))
+                    feed_is_live.set()
+
+                    async for typ, msg in msg_gen:
+                        await send_chan.send({sym: msg})
+
+
+@acm
+async def subscribe(ws: wsproto.WSConnection, connect_id, sym):
```
goodboy, line 559: looks wayyy bettter

piker/brokers/kucoin.py
```diff
@@ -0,0 +1,718 @@
...
+                            'price_tick_size': float(pair.baseIncrement),
+                            'lot_tick_size': float(pair.baseMinSize),
+                        },
+                        'shm_write_opts': {'sum_tick_vml': False},
+                        'fqsn': sym,
+                    }
+                }
+
+                async with (
+                    subscribe(ws, connect_id, kucoin_sym),
```
goodboy, line 542: FYI, you can also pass this into `open_autorecon_ws(ws_url, fixture=subscribe) as ws,` above. Again nbd we can polish it later especially since that API has been reworked entirely in #489 😂

goodboy (approved):
@jaredgoldman other then a couple more suggestions that I don't think you need to change right now in this PR (particularly bc i just had to entirely rework `open_autiorecon_ws()` in #489 😂) and this little bit of linting i was still able to catch:

screenshot-2023-04-21_14-11-19

I think this is more then ready to land!

Great work yet again and glad to see our backend crypto set expanding using `trio` native machnery 🏄🏼 

**Only last thing** is please move all the outstanding todo tasks and follow up into another issue that links to this one **before you merge this**.

Other then that please merge at your leisure 🐶

piker/brokers/kucoin.py
jaredgoldman: Pattern matching against subject doesn't work for some reason after boxing it with the msgspec struct. I've added a default case with a console.warn log for unexpected messages

piker/brokers/kucoin.py
jaredgoldman: Ah but then I wouldn't have my symbol data.

jaredgoldman:
@goodboy https://github.com/pikers/piker/issues/499

goodboy:
@jaredgoldman :boom:

excellent work yet again compadre 🏄🏼 

nice follow up issue and final linting before getting this in ma dawg!

PR #497, opened 2023-04-11
Only emit `'trade'` ticks on msgs with new(er) timestamps

goodboy:
Not sure it needs more explanation but, more or less `kucoin` is super weird with emitting msgs without actual data updates?

I've solved it by only emitting "trade" ticks on ts increments.
I wonder if we can find a sane way to log some diffs in these messages to console so we can see what's actually changing?


goodboy:
Missed a commit before a submitting this.

All covered in #494

PR #506, opened 2023-05-04
`python3.11` support... ish.

goodboy:
python 3.11 just landed on arch 🏄🏼 

Everything seems to be running fine for me with the small `numpy.float` removal 💥 

Please give it a shot if you've got a copy of the new cpython interpreter 😎 

----
#### wut's new bby!
https://docs.python.org/3/whatsnew/3.11.html

### DIFF
--- a/piker/data/_sharedmem.py
+++ b/piker/data/_sharedmem.py
@@ -271,7 +271,7 @@ class ShmArray:
 
         # type that all field values will be cast to
         # in the returned view.
-        common_dtype: np.dtype = np.float,
+        common_dtype: np.dtype = float,
 
     ) -> np.ndarray:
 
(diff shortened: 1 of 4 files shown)


goodboy:
Oof! stuff still a bit borked for `pikerd --tsdb`, namely docker py apis that changed; patch comin stat!

goodboy:
I think this is also due to 3.11 which is causing an `asyncio` breakage inside `ib_insync`:
https://github.com/erdewit/ib_insync/issues/585

goodboy (commented):
I mean it's close enough minus backends that are being finicky..

goodboy:
Hah! figured out the issue finally..

Turns out they removed support for passing coros to `asyncio.wait()`..
https://docs.python.org/3/library/asyncio-task.html#asyncio.wait

So we need to instead use `.wait_for()` on the `ib_insync.Ticker.TickerUpdateEvent: eventkit.Event`..

PR #520, opened 2023-06-09
Binancial secs: futes for the peepz!

goodboy:
Replacement for #182 keeping as much of the work from @guilledk as possible!

This is the first proto which will begin to complete the tasks listed in #519 🏄🏼 

More then likely I am going to rebase this onto #486 as well since I'm somewhat hoping that lands before this onto mainline.

ping @ebisu4
---
#### Summary:
- [x] adds first proto support for futures testnet endpoints, rest API and drafted functionality to support live order control via our EMS:
  ```python
  _url = 'https://api.binance.com'
  _sapi_url = 'https://api.binance.com'
  _fapi_url = 'https://testnet.binancefuture.com'
  ```
  - thanks @guilledk for all that work from 2 years ago 😂 
- [ ] futures live data feeds via ws api as per requirements in #519
  - [ ] paper engine trading using these feeds alongside spot feeds
- [ ] ideally basic limit order control through `emsd` with accompanying tests

More follow up (smaller) PRs will likely come to address the more advanced features desired in #519

PR #525, opened 2023-06-20
`msgspec` struct derivative updates

goodboy:
Ran into some weird `.data.types.Struct.__annotations__` behavior while workin on #520 so ended up digging and finding the new `msgspec.structs` API(s) and instead re-implemented our type-casting methods using that.

More or less just a super simple delegation patch in the end 🥳


ebisu4 (approved):
Looks good to me

PR #526, opened 2023-06-27
Basic buy bot

goodboy:
Finally a first rough draft of an automatic trading *bot* using multiple subsystems throughout the codebase 🏄🏼 

Currently this example just loads an order client and live data feeds and then submits a trailing buy order that shouldn't clear:
- presuming ems handling is fast enough; keeps the price 0.004% below the last clearing price. 
- uses the internal paper engine such that order updates are more or less as low latency as they can be given we have no serious clearing model (yet) inside.

---
#### TODO:
- [ ] we need some wrapping to make the order control API higher level and to hide many of the runtime boot details:

.. write this up ..


goodboy:
bleh, those last 2 commits needs to be squashed obviously 😂

PR #531, opened 2023-07-18
add tests for brokers.toml file

di1ara:
added auto file gen for brokers.toml and tests

### DIFF
--- a/piker/config.py
+++ b/piker/config.py
@@ -427,7 +427,7 @@ def load_accounts(
 
 ) -> bidict[str, str | None]:
 
-    conf, path = load()
+    conf, path = load(touch_if_dne=True)
     accounts = bidict()
     for provider_name, section in conf.items():
         accounts_section = section.get('accounts')
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -174,13 +174,14 @@ def tmpconfdir(
     )
     assert path.is_file(), 'WTH.. `brokers.toml` not created!?'
 
-    return tmpconfdir
+    yield tmpconfdir
 
     # NOTE: the `tmp_dir` fixture will wipe any files older then 3 test
     # sessions by default:
     # https://docs.pytest.org/en/6.2.x/tmpdir.html#the-default-base-temporary-directory
     # BUT, if we wanted to always wipe conf dir and all contained files,
-    # rmtree(str(tmp_path))
+    from shutil import rmtree
+    rmtree(str(tmp_path))
 
 
 @pytest.fixture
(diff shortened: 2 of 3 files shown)


tests/conftest.py
```diff
@@ -174,13 +174,14 @@ def tmpconfdir(
...
 
-    return tmpconfdir
+    yield tmpconfdir
 
     # NOTE: the `tmp_dir` fixture will wipe any files older then 3 test
     # sessions by default:
     # https://docs.pytest.org/en/6.2.x/tmpdir.html#the-default-base-temporary-directory
     # BUT, if we wanted to always wipe conf dir and all contained files,
-    # rmtree(str(tmp_path))
+    from shutil import rmtree
```
goodboy, line 183: Hmm, my one question was whether we ended up needing this?

I can't remember if we are auto-injecting the per-test user config tmpdir or it stays pseudo-global in between each test scope.

Not a huge deal for now (since explicit cleanup is normally always better).

di1ara: I'm not sure why we added this. but I commented it and tested, it still working there is no error. maybe we don't need it

goodboy: Yeah, so i think that means the tests are isolated in terms of the user-dir created per test. So yeah, leave it commented as it was for now 👍🏼

tests/test_configs.py
```diff
@@ -0,0 +1,10 @@
+from pathlib import Path
+
+from piker import config
+
+
+def test_root_conf_networking_section():
```
goodboy, line 6: Ok so obviously we need to change the name of this 😂 

I would change it to something like `test_brokersconfig_created_created_from_template_on_load_accounts()`

or something..

goodboy: I'd also add a doc string explaining that this issue was found when trying to run `piker chart` on windows.

some resources to help you:
- https://peps.python.org/pep-0257/#what-is-a-docstring
- https://peps.python.org/pep-0008/#documentation-strings

Also, you can look at other tests which have them to get a rough idea.

piker/config.py
```diff
@@ -427,7 +427,7 @@ def load_accounts(
 
 ) -> bidict[str, str | None]:
 
-    conf, path = load()
+    conf, path = load(touch_if_dne=True)
```
goodboy, line 430: Longer term I wonder if we should change the name of this flag to something like `cpy_from_template_if_dne: bool` or something.

**Not** something we have to change in this PR, just rubber ducking.

goodboy (approved):
Looks good to me 😎 

@di1ara thanks again!

guilledk (approved):

PR #533, opened 2023-08-14
Poetry2nix

goodboy:
First draft hopefully getting `poetry` pkging and `poetry2nix` support working!


goodboy:
Sup all, I think this is finally nearly ready for testing and reviewing!

So far there's still no `--with uis` support on `nix` but I may just try to land the poetry pkging first and follow up to solve that stuff since it's going to require some patches to nixpkgs and possibly `poetry2nix` itself 😂 

I will be updating the readme and adding a proper todo list before merge to the description above 👍🏼

goodboy:
Way out of date, see `gitea:main`.

We now use an impure template from `pyproject.nix` and/or will be moving to use `uv2nix` when/if needed.

PR #537, opened 2024-01-17
Fix `develop.nix` file for poetry install

guilledk:


guilledk:
Dropped in favour of #538

